Issue #95, opened 2016-03-30
Switch inclusion of header files from <> to ""


kevyang:

elfring:
Would you like to [reconsider this proposal](http://stackoverflow.com/questions/21593/what-is-the-difference-between-include-filename-and-include-filename "What is the difference between #include <filename> and #include “filename”?")?

thinkingfish:
@elfring Can you elaborate? 

To clarify: We currently almost use always use `<>`, this issue is suggesting using `""` for all the headers local within the pelikan repo (i.e. excluding deps – `ccommon` and systems headers)

elfring:
Do the include statements which are using angle brackets so far refer to header files that belong to application programming interfaces?

thinkingfish:
@elfring Yes. We basically _were_ using angle brackets for _everything_.

elfring:
Is [the use of angle brackets](http://stackoverflow.com/questions/21593/what-is-the-difference-between-include-filename-and-include-filename "What is the difference between #include <filename> and #include “filename”?") more appropriate occasionally?

thinkingfish:
@elfring I'm not exactly sure what the best practice is. I've seen people using angle brackets *only* for system libraries, and I've seen the other extreme where angle brackets are used liberally (twemcache, twemproxy actually fell into the latter camp).

For now, we have restricted angle brackets to be used for anything that's *not* under `pelikan/src`, including `ccommon` includes, which seems to a little more liberal than average. This is mostly to make local includes – such as having the `.c` include the corresponding `.h` – as concise as possible.

If anybody knows more feel free to comment.

PR #110, opened 2016-04-19
include pelikan header files with double quotes instead of angle brackets

kevyang:
issue #95 

everything compiles and links correctly

### DIFF
--- a/src/server/pingserver/admin/process.c
+++ b/src/server/pingserver/admin/process.c
@@ -1,7 +1,7 @@
 #include "process.h"
 
-#include "../../../protocol/admin/admin_include.h"
-#include "../../../util/procinfo.h"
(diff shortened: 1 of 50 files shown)


src/server/pingserver/admin/process.c
```diff
@@ -1,7 +1,7 @@
 #include "process.h"
 
-#include <protocol/admin/admin_include.h>
-#include <util/procinfo.h>
+#include "../../../protocol/admin/admin_include.h"
```
thinkingfish: All these are fairly ugly.
You can drop the leading `../../../` and it would still work. In this case it will just fall back to search in the paths specified by `-I`, and we have added the top-level `src/` to the include directories.

thinkingfish:
:ship: feel free to merge.

PR #113, opened 2016-05-04
declare project as C-only to avoid looking for CXX related symbols

thinkingfish:
This reduces the dependency requirement when building docker images (otherwise we will have to pull g++ when building image based on debian:jessie)


seppo0010:
lgtm, travis seems to have a spurious failure in a wheel test, which is not surprising because those are time dependent and travis environments are known for being slow

thinkingfish:
@seppo0010 yea, I think eventually we can write a mock-timer and only run against the real clock in a more predictable environment.

Issue #116, opened 2016-05-23
de-register writable event on a socket after successful `_worker_event_write()`


thinkingfish:
Currently write/writable event for a socket is registered only when a call of `write` would block for the corresponding `fd`, however, once added, this event is never removed until all events on the `fd` are, usually upon to connection termination.

This behavior is incorrect, at least for our epoll implementation. Since we do not want to simply add the `EPOLLONESHOT` flag to `event_add_write`, which disables all future events upon firing (and the event fired could be either read or write, so that's not particularly useful), we have a couple of options:
- disable read and add writable event to the `fd` with `EPOLLONESHOT`, and re-add read only when the next write is successful. This implies we will not try to process more data coming from the socket until we have flushed out previous responses for this connection;
- add return status to `_worker_event_write()`, and upon a successful write, modify the events associated with the filedescriptor to remove write-related flags.

PR #117, opened 2016-07-02
disable EVENT_READ when waiting for EVENT_WRITE, and vice versa

thinkingfish:
This addresses https://github.com/twitter/pelikan/issues/116

We are wiating only for either read or write event at any given point, so difficulty to write to socket will translate (eventually) to back pressure by ignoring incoming data from the same channel.


kevyang:
ship it

Issue #119, opened 2016-09-27
binaries created in source directory


lomew:
The `pelikan_twemcache` and other binaries are left in `SRCDIR/_bin` rather than `BUILDDIR/_bin`, as would be expected and what the documentation describes.

PR #120, opened 2016-09-27
build: do not put binaries into source dir (#119)

lomew:
Build was putting binaries, such as pelikan_twemcache, into
CMAKE_HOME_DIRECTORY/_bin, which is the source directory.

Fixed to use PROJECT_BINARY_DIR.  This also works if we are within another
project via add_subdirectory(), our binaries will land at the top of the
pelikan build directory, not up in the enclosing project.


thinkingfish:
@lomew Sorry for overlooking this while I was on leave. Feel free to merge such changes yourself next time.

PR #123, opened 2017-01-21
Add data structures, interface, and basic unit tests for hotkey sampl…

kevyang:
…ing/detection


src/hotkey/counter_table.c
```diff
@@ -0,0 +1,242 @@
...
+#include <cc_mm.h>
+#include <cc_pool.h>
+
+#define COUNTER_TABLE_MODULE_NAME "hotkey::counter_table"
+
+struct counter_table_entry {
+    STAILQ_ENTRY(counter_table_entry) next; /* entry in hash table or pool */
+
+    char                              key[MAX_KEY_LEN];
+    uint32_t                          nkey;
```
thinkingfish: klen might be a better name?

src/hotkey/counter_table.c
```diff
@@ -0,0 +1,242 @@
+#include "counter_table.h"
```
thinkingfish: Sorry for the possible bike-shedding on naming. But counter_table doesn't seem the most intuitive to me. May I propose "key-count map" or "kc_map" as an alternative?

src/hotkey/counter_table.c
```diff
@@ -0,0 +1,242 @@
...
+}
+
+static inline struct cte_slh *
+_get_bucket(const char *key, uint32_t nkey)
+{
+    return &(table[hash(key, nkey, 0) % table_size]);
+}
+
+uint32_t
+counter_table_incr(char *key, uint32_t nkey)
```
thinkingfish: use ```struct bstring``` to pass around ```key```?

src/hotkey/hotkey.h
```diff
@@ -0,0 +1,36 @@
...
+#include <cc_option.h>
+
+#include <stdbool.h>
+#include <stdint.h>
+
+/* TODO(kevyang): add stats for hotkey module */
+
+#define HOTKEY_NSAMPLE   10000 /* keep last 10000 keys sampled by default */
+#define HOTKEY_RATE      100   /* sample one in every 100 keys by default */
+#define HOTKEY_THRESHOLD 10    /* signal for hotkey if 10 or more keys in sample by default */
```
thinkingfish: Wonder if it makes more sense to use a percentage, so we don't always have to tune HOTKEY_NSAMPLE & HOTKEY_THRESHOLD at the same time. E.g.
```
HOTKEY_THRESHOLD_PCT 0.01
HOTKEY_THRESHOLD (unsigned)(HOTKEY_NSAMPLE * HOTKEY_THREHOLD_PCT)
```

src/hotkey/hotkey.h
```diff
@@ -0,0 +1,36 @@
+#pragma once
+
+#include <cc_option.h>
+
+#include <stdbool.h>
+#include <stdint.h>
+
+/* TODO(kevyang): add stats for hotkey module */
+
+#define HOTKEY_NSAMPLE   10000 /* keep last 10000 keys sampled by default */
```
thinkingfish: More on naming, ```WINDOW_SIZE```?

src/hotkey/hotkey.h
```diff
@@ -0,0 +1,36 @@
...
+} hotkey_options_st;
+
+extern bool hotkey_enabled;
+
+void hotkey_setup(hotkey_options_st *options);
+void hotkey_teardown(void);
+
+#define hotkey_sample(key, nkey) do { \
+    if (hotkey_enabled) {                            \
+        _hotkey_sample(key, nkey);                   \
```
thinkingfish: macro probably an overkill here. We are unlikely to call it more than once per request, so that's not too bad. Different story elsewhere we used this tactic (logging, metrics)

src/hotkey/queue.c
```diff
@@ -0,0 +1,190 @@
+#include "queue.h"
```
thinkingfish: queue a bit too generic given we have a cc_queue in ccommon

Can we call it something more context-aware such as key_window or something (not thinking very hard here right now)

src/hotkey/queue.c
```diff
@@ -0,0 +1,190 @@
...
+
+static bool queue_init = false;
+static int queue_size = 0;
+
+FREEPOOL(qn_pool, qnq, queue_node);
+static struct qn_pool qnp;
+static bool qnp_init = false;
+
+struct queue_node {
+    char                     key[MAX_KEY_LEN];
```
thinkingfish: One thought- most of the storage used here are for storing keys, and we are storing it (at most) twice, once in the queue/window, once in the key-counter map. However, one property of the data structures is that every key in the map will be in the queue at at least one location. Therefore, it should be possible to use a pointer in the hash table/map to point to the key as stored in the queue. True it's not too big a deal (~ a few MBs) given the number of samples we stored, but I kinda like the idea.

src/hotkey/counter_table.c
```diff
@@ -0,0 +1,242 @@
...
+    *counter_table_entry = NULL;
+}
+
+static void
+counter_table_entry_pool_destroy(void)
+{
+    struct counter_table_entry *cte, *tcte;
+
+    if (!ctep_init) {
+        log_warn("counter_table_entry pool was not created, ignore");
```
seppo0010: missing a `return;`?

src/hotkey/hotkey.c
```diff
@@ -0,0 +1,79 @@
...
+    hotkey_init = true;
+}
+
+void
+hotkey_teardown(void)
+{
+    log_info("Tear down the %s module", HOTKEY_MODULE_NAME);
+
+    if (!hotkey_init) {
+        log_warn("%s was not setup", HOTKEY_MODULE_NAME);
```
seppo0010: missing `return;`?

src/hotkey/queue.c
```diff
@@ -0,0 +1,190 @@
...
+    *queue_node = NULL;
+}
+
+static void
+queue_node_pool_destroy(void)
+{
+    struct queue_node *qn, *tqn;
+
+    if (!qnp_init) {
+        log_warn("queue_node pool was not created, ignore");
```
seppo0010: once again

src/hotkey/queue.c
```diff
@@ -0,0 +1,190 @@
...
+
+void
+queue_teardown(void)
+{
+    struct queue_node *qn, *tqn;
+
+    log_info("tear down the %s module", QUEUE_MODULE_NAME);
+
+    if (!queue_init) {
+        log_warn("%s was not setup", QUEUE_MODULE_NAME);
```
seppo0010: here too

test/hotkey/counter_table/check_counter_table.c
```diff
@@ -0,0 +1,99 @@
...
+}
+
+/**************
+ * test cases *
+ **************/
+
+START_TEST(test_basic)
+{
+    char *key1 = "key1", *key2 = "key2";
+    uint32_t klen = 4;
```
seppo0010: Use two keys with different length. Also, since the char* is actually 5 bytes as it has a NULL termination, use 5 instead of 4 in this case, as it may help catch off-by-one errors.

src/hotkey/queue.c
```diff
@@ -0,0 +1,190 @@
...
+    qn->nkey = nkey;
+    STAILQ_INSERT_TAIL(&q, qn, next);
+    ++queue_size;
+}
+
+uint32_t
+queue_pop(char *buf)
+{
+    struct queue_node *qn = STAILQ_FIRST(&q);
+    uint32_t nkey;
```
seppo0010: `ASSERT(queue_size);`?

test/hotkey/queue/check_queue.c
```diff
@@ -0,0 +1,133 @@
...
+    ck_assert_int_eq(queue_len(), 0);
+    ck_assert_int_eq(queue_pop_len, klen);
+    ck_assert(strncmp(key, buf, klen) == 0);
+}
+END_TEST
+
+START_TEST(test_multiple)
+{
+    char *key1 = "key1", *key2 = "key22", *key3 = "key333";
+    uint32_t klen1 = 4, klen2 = 5, klen3 = 6;
```
seppo0010: 5, 6 and 7?

src/hotkey/queue.c
kevyang: I will do this in a separate pull request. This is a larger change than it seems, as in order to do this we need to make the queue aware of the counting map

test/hotkey/counter_table/check_counter_table.c
kevyang: I made the sizes the length of the strings without null termination, because we don't use null terminated keys in cache.

src/hotkey/key_window.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include <stdint.h>
+
+/*
+ * The key_window module provides a FIFO key_window interface to help with bookkeeping the
```
thinkingfish: Given that we don't really update the size of key window, does it make things easier to use ring array to implement this data structure?

test/hotkey/key_window/check_key_window.c
```diff
@@ -0,0 +1,145 @@
...
+
+    ck_assert_int_eq(key_window_len(), 0);
+    ck_assert_int_eq(key_window_pop_len, key1.len);
+    ck_assert(strncmp(key1.data, buf, key1.len) == 0);
+}
+END_TEST
+
+START_TEST(test_multiple)
+{
+    char *key1str = "key1", *key2str = "key22", *key3str = "key333";
```
thinkingfish: use macros to define key1/2/3, as we did in other unit tests? (see protocol/data/memcache)

Also, you can use str2bstr to convert a C string to a struct bstring. https://github.com/twitter/ccommon/blob/master/include/cc_bstring.h#L38

src/hotkey/key_window.h
kevyang: I think the benefits of using ring_array do not significantly outweigh the drawbacks. Although it would provide a clean push/pop, using STAILQ is also a one liner in this aspect and is not significantly harder to understand. ring_array also lacks a foreach equivalent, which means we would need to write a messier loop for freeing the contents. Additionally, using ring_array still does not remove the need to check whether the the window is full. Furthermore, if we are to port this hotkey change to twemcache, twemcache already has mc_queue so it would be mostly read to go; however, were we to switch to ring_array we would need to port ring_array over to twemcache as well. For these reasons, I think we should keep it as it is for now.

thinkingfish: Good point on the lack of foreach in ring array.
The tradeoff makes sense for now.

thinkingfish: If we can preallocate the entire window when initialize, can we just do that, treat the whole thing as a big array, use offset to traverse the structure and forget about the link list?

PR #124, opened 2017-01-24
make twemcache deal with partial request value

thinkingfish:
This is the first diff to address issue #122 

Subsequent diff will add more unit tests, as well as handle partial value when writing responses (when wbuf is full)

### DIFF
--- a/src/protocol/data/memcache/parse.c
+++ b/src/protocol/data/memcache/parse.c
@@ -287,7 +287,7 @@ _parse_val(struct bstring *val, struct buf *buf, uint32_t nbyte)
     }
 
     val->len = MIN(nbyte, rsize);
-    val->data = buf->rpos;
+    val->data = (val->len > 0) ? buf->rpos : NULL;
     buf->rpos += val->len;
 
     /* verify CRLF */
@@ -866,6 +866,7 @@ parse_req(struct request *req, struct buf *buf)
                     status = PARSE_OK;
                     req->rstate = REQ_PARTIAL;
                 } else {
+                    ASSERT(req->first == 1);
                     log_verb("try to left shift a request when possible");
                     request_reset(req);
                     buf->rpos = old_rpos;
--- a/src/protocol/data/memcache/parse.c
+++ b/src/protocol/data/memcache/parse.c
@@ -273,27 +273,32 @@ _chase_uint(uint64_t *num, struct buf *buf, bool *end, uint64_t max)
 }
 
 static parse_rstatus_t
-_parse_val(struct bstring *val, struct buf *buf, uint32_t vlen)
+_parse_val(struct bstring *val, struct buf *buf, uint32_t nbyte)
 {
     parse_rstatus_t status;
+    uint32_t rsize = buf_rsize(buf);
+    bool complete = true;
 
     log_verb("parsing val (string) at %p", buf->rpos);
 
-    if (buf_rsize(buf) < vlen + CRLF_LEN) {
-        log_verb("buf %p has %"PRIu32" out of the %"PRIu32" bytes expected",
-                buf, buf_rsize(buf), vlen + CRLF_LEN);
+    if (rsize < nbyte + CRLF_LEN) {
+        complete = false;
+        status = PARSE_EUNFIN;
 
-        return PARSE_EUNFIN;
+        log_verb("buf %p has %"PRIu32" out of the %"PRIu32" bytes expected",
+                buf, rsize, nbyte + CRLF_LEN);
     }
 
-    val->len = vlen;
+    val->len = MIN(nbyte, rsize);
     val->data = buf->rpos;
-    buf->rpos += vlen;
+    buf->rpos += val->len;
 
     /* verify CRLF */
-    status = _try_crlf(buf, buf->rpos);
-    if (status == PARSE_OK) {
-        buf->rpos += CRLF_LEN;
+    if (complete) { /* _try_crlf cannot return EUNFIN in this case */
+        status = _try_crlf(buf, buf->rpos);
+        if (status == PARSE_OK) {
+            buf->rpos += CRLF_LEN;
+        }
     }
 
     return status;
@@ -657,6 +662,7 @@ _subrequest_store(struct request *req, struct buf *buf, bool *end, bool cas)
         return status;
     }
     req->vlen = (uint32_t)n;
+    req->nremain = req->vlen;
     /* CAS, conditional */
     if (cas) {
         if (*end) {
@@ -782,31 +788,81 @@ _parse_req_hdr(struct request *req, struct buf *buf)
 parse_rstatus_t
 parse_req(struct request *req, struct buf *buf)
 {
-    parse_rstatus_t status = PARSE_EUNFIN;
+    parse_rstatus_t status = PARSE_OK;
     char *old_rpos = buf->rpos;
+    bool leftmost = (buf->rpos == buf->begin);
 
-    ASSERT(req->rstate == REQ_PARSING);
+    /*
+     * we allow partial value in the request (but not the head portion),
+     * so that we can incrementally fill in a large value over multiple socket
+     * reads. This is more useful for the server which allows more predictable
+     * buffer management (e.g. no unbounded read buffer). Currently partial
+     * value is not implemented for the response.
+     */
+    switch (req->rstate) {
+    case REQ_PARSING: /* a new request */
+        log_verb("parsing buf %p into req %p", buf, req);
+        status = _parse_req_hdr(req, buf);
+        if (status == PARSE_EUNFIN) {
+            log_verb("incomplete data: reset read position, jump back %zu bytes",
+                    buf->rpos - old_rpos);
+            /* return and start from beginning next time */
+            buf->rpos = old_rpos;
+            return PARSE_EUNFIN;
+        }
+        /* fall-through intended */
 
-    log_verb("parsing buf %p into req %p", buf, req);
+    case REQ_PARTIAL: /* continuation of value parsing for the current request */
+        if (status == PARSE_OK && req->val) {
+            status = _parse_val(&req->vstr, buf, req->nremain);
+        }
+        if (status == PARSE_OK) {
+            req->rstate = REQ_PARSED;
+            req->partial = false;
+            INCR(parse_req_metrics, request_parse);
+        } else {
+            if (status != PARSE_EUNFIN) {
+                log_debug("parse req returned error state %d", status);
+                req->cerror = 1;
+                INCR(parse_req_metrics, request_parse_ex);
+            } else { /* partial val, we return upon partial header above */
+                /*
+                 * We try to fit as much data into read buffer as possible
+                 * before processing starts. When request starts somewhere in
+                 * the middle of buf, we jump back and wait for more data to
+                 * arrive (and expect caller to left-shift data in buf)
+                 *
+                 * This is a seemingly weird and unnecessary decision, the
+                 * reason we need to do this is because we want to allow
+                 * partial value only for set/add/cas/replace, but not for
+                 * append/prepend. Because append/prepend are modifying keys
+                 * already linked into hash, if we want to support partial
+                 * value we need to either copy the key/value or temporarily
+                 * unlink the key. And either option has severe drawbacks. Given
+                 * append/prepend very large value is a case that I've never
+                 * seen any need for in the field, it's a reasonable assumption
+                 * to make, at least for now.
+                 *
+                 * With this behavior in place, the processing logic can assume
+                 * that if it sees a partial request for append/prepend, the
+                 * payload is too big to be held in the read buffer, without the
+                 * possibility that a small append request just happens to come
+                 * behind a number of other requests.
+                 */
+                if (leftmost) {
+                    req->partial = true;
+                    status = PARSE_OK;
+                    req->rstate = REQ_PARTIAL;
+                } else {
+                    buf->rpos = old_rpos;
+                }
+            }
+        }
+        break;
 
-    status = _parse_req_hdr(req, buf);
-    if (status == PARSE_OK && req->val) {
-        status = _parse_val(&req->vstr, buf, req->vlen);
-    }
-
-    if (status == PARSE_EUNFIN) {
-        log_verb("incomplete data: reset read position, jump back %zu bytes",
-                buf->rpos - old_rpos);
-        buf->rpos = old_rpos; /* start from beginning next time */
-        return PARSE_EUNFIN;
-    }
-    if (status != PARSE_OK) {
-        log_debug("parse req returned error state %d", status);
-        req->cerror = 1;
-        INCR(parse_req_metrics, request_parse_ex);
-    } else {
-        req->rstate = REQ_PARSED;
-        INCR(parse_req_metrics, request_parse);
+    default:
+        NOT_REACHED();
+        status = PARSE_EOTHER;
     }
 
     return status;
--- a/src/protocol/data/memcache/parse.c
+++ b/src/protocol/data/memcache/parse.c
@@ -277,21 +277,17 @@ _parse_val(struct bstring *val, struct buf *buf, uint32_t nbyte)
 {
     parse_rstatus_t status;
     uint32_t rsize = buf_rsize(buf);
-    bool complete = true;
 
     log_verb("parsing val (string) at %p", buf->rpos);
-
-    if (rsize < nbyte + CRLF_LEN) {
-        complete = false;
-        status = PARSE_EUNFIN;
-    }
 
     val->len = MIN(nbyte, rsize);
     val->data = (val->len > 0) ? buf->rpos : NULL;
     buf->rpos += val->len;
 
     /* verify CRLF */
-    if (complete) { /* _try_crlf cannot return EUNFIN in this case */
+    if (rsize < nbyte + CRLF_LEN) {
+        status = PARSE_EUNFIN;
+    } else {
         status = _try_crlf(buf, buf->rpos);
         if (status == PARSE_OK) {
             buf->rpos += CRLF_LEN;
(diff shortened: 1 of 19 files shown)


test/storage/slab/check_slab.c
```diff
@@ -116,6 +113,96 @@ START_TEST(test_insert_large)
...
+
+    /* backfill */
+    val.len = vlen - val.len;
+    val.data = cc_alloc(val.len);
+    cc_memset(val.data, 'B', val.len);
+    item_backfill(it, &val, false);
+    free(val.data);
+    ck_assert_msg(!it->is_linked, "item linked by mistake");
+    ck_assert_int_eq(it->vlen, vlen);
+    ck_assert_msg(strspn(item_data(it) + vlen - val.len, "B") == val.len,
```
seppo0010: add an assertion on the first part of the item_data to make sure it was not modified?

test/storage/slab/check_slab.c
```diff
@@ -492,6 +548,80 @@ START_TEST(test_evict_lru_basic)
...
+    val = null_bstring;
+    item_backfill(it, &val, false);
+    ck_assert_msg(s->refcount == 1, "slab refcount %"PRIu32"; 1 expected", s->refcount);
+    val = null_bstring;
+    item_backfill(it, &val, true);
+    ck_assert_msg(s->refcount == 0, "slab refcount %"PRIu32"; 0 expected", s->refcount);
+}
+END_TEST
+
+START_TEST(test_evict_refcount)
```
seppo0010: isn't this test exactly the same as the one above?

src/storage/slab/slab.h
```diff
@@ -105,6 +107,7 @@ struct slab {
     rel_time_t        utime;        /* last update time in secs */
     uint8_t           id;           /* slabclass id */
     uint32_t          padding:24;   /* unused */
+    uint32_t          refcount;     /* number of reserved items */
```
seppo0010: I don't think this description is clear, it makes me thing that each slab reserves and not that it is itself reserved

src/storage/slab/slab.h
```diff
@@ -62,13 +62,15 @@ typedef struct {
...
-    ACTION( item_insert,        METRIC_COUNTER, "# items inserted"         )\
-    ACTION( item_remove,        METRIC_COUNTER, "# items removed"          )
+    ACTION( item_alloc,         METRIC_COUNTER, "# items allocated"        )\
+    ACTION( item_alloc_ex,      METRIC_COUNTER, "# item alloc errors"      )\
+    ACTION( item_dealloc,       METRIC_COUNTER, "# items de-allocated"     )\
+    ACTION( item_linked_curr,   METRIC_GAUGE,   "# current items, linked"  )\
+    ACTION( item_link,          METRIC_COUNTER, "# items inserted to HT"   )\
+    ACTION( item_unlink,        METRIC_COUNTER, "# items removed from HT"  )\
+    ACTION( item_keyval_byte,   METRIC_GAUGE,   "key+val in bytes, linked" )\
+    ACTION( item_val_byte,      METRIC_GAUGE,   "value only in bytes"      )
```
seppo0010: I feel there should be a gauge for the items that are backfilling, reserved and not linked... maybe `alloc`-`link` already provides that?

test/storage/slab/check_slab.c
```diff
@@ -116,6 +113,96 @@ START_TEST(test_insert_large)
...
+    /* backfill & link */
+    val.len = 0;
+    item_backfill(it, &val, true);
+    ck_assert_msg(it->is_linked, "completely backfilled item not linked");
+    ck_assert_int_eq(it->vlen, VLEN);
+    ck_assert_msg(strspn(item_data(it), "A") == VLEN,
+            "item_data contains wrong value %.*s", VLEN, item_data(it));
+
+    /* release */
+    item_release(&it);
```
seppo0010: `it` is linked, should not use `item_release` according to comment above

src/server/twemcache/data/process.c
```diff
@@ -532,7 +571,7 @@ twemcache_process_read(struct buf **rbuf, struct buf **wbuf, void **data)
 
         status = parse_req(req, *rbuf);
         if (status == PARSE_EUNFIN) {
-            goto done;
+            return 0;
```
seppo0010: I don't fully understand these changes of `return`ing instead of `goto`... `_cleanup` is no longer called, that seems to be on purpose, but I don't understand why

src/server/twemcache/data/process.c
```diff
@@ -225,12 +251,12 @@ _process_replace(struct response *rsp, struct request *req)
...
     struct bstring *key;
+    struct item *it;
 
     INCR(process_metrics, replace);
     key = array_first(req->keys);
-    if (item_get(key) != NULL) {
-        item_delete(key);
-        status = item_insert(key, &(req->vstr), req->flag, time_reltime(req->expiry));
+    if ((it = item_get(key)) != NULL) {
+        status =_generic_set(req, true);
```
seppo0010: missing space after `=`

src/server/twemcache/data/process.c
```diff
@@ -262,8 +288,7 @@ _process_cas(struct response *rsp, struct request *req)
         rsp->type = RSP_EXISTS;
         INCR(process_metrics, cas_exists);
     } else {
-        item_delete(key);
-        status = item_insert(key, &(req->vstr), req->flag, time_reltime(req->expiry));
+        status =_generic_set(req, true);
```
seppo0010: missing space after `=`

src/server/twemcache/data/process.c
```diff
@@ -207,7 +233,7 @@ _process_add(struct response *rsp, struct request *req)
         rsp->type = RSP_NOT_STORED;
         INCR(process_metrics, add_notstored);
     } else {
-        status = item_insert(key, &(req->vstr), req->flag, time_reltime(req->expiry));
+        status =_generic_set(req, false);
```
seppo0010: missing space after `=`

src/server/twemcache/data/process.c
```diff
@@ -174,16 +174,42 @@ _error_rsp(struct response *rsp, item_rstatus_t status)
...
 {
     item_rstatus_t status;
-    struct bstring *key;
-
     INCR(process_metrics, set);
-    key = array_first(req->keys);
-    item_delete(key);
-    status = item_insert(key, &(req->vstr), req->flag, time_reltime(req->expiry));
+
+    status =_generic_set(req, true);
```
seppo0010: missing space after `=`

src/protocol/data/memcache/request.c
```diff
@@ -39,6 +39,10 @@ request_reset(struct request *req)
     req->delta = 0;
     req->vcas = 0;
 
+    req->partial = false; /* only set to true when partial value received */
```
seppo0010: having a `rstate` that may be `REQ_PARTIAL` this sees either redundant, or at least confusing... looking at the change I can see partial being set to `false` only when `rstate` going from `REQ_PARTIAL` to `REQ_PARSED` and set to `true` only when rstate is set to `REQ_PARTIAL`... isn't then this boolean redundant?

test/storage/slab/check_slab.c
```diff
@@ -116,6 +113,96 @@ START_TEST(test_insert_large)
...
+    item_rstatus_t status;
+    struct item *it;
+    uint32_t vlen, dataflag = 12345;
+
+    test_reset();
+
+    key = str2bstr(KEY);
+
+    vlen = 1000 * KiB;
+    val.len = vlen / 2;
```
thinkingfish: Good point, I'll update it in the next commit.

src/storage/slab/slab.h
thinkingfish: updated in the next commit, please see if it's better

" # items that can't be evicted"

src/storage/slab/slab.h
thinkingfish: Yea it does. I wonder if this is a metric that's gonna be useful under normal operations. I feel people may care about it only when they receive a OOM (due to failure in eviction), which should be rare.

test/storage/slab/check_slab.c
thinkingfish: This test changes the slab config so the second reserve should fail, instead of succeed, (ITEM_ENOMEM expected). And another reserve is attempted after ```item_insert```, which should succeed now that we can evict.

thinkingfish:
@seppo0010 To your comment on having a different type of tests, I agree 100%. I think you are referring to integration testing or validation testing. Basically, we want to have a reasonable set of request-response pairs that we want to make sure always work against one or more server configurations.

For this particular PR, I manually tested many simple commands (all the write commands in memcached, in various orders, w/ race, etc). But clearly I have not captured or automated them yet.

I am not sure what's the best tool to do that, short of writing/importing a full-fledged memcached client and use another scripting language to define the test sequence. I'll be grateful if you or someone can shed some light on this.

We may also consider putting such functional tests in a separate repo, since I am slightly concerned about too many dependencies on building the binaries.

seppo0010:
We can have test dependencies that are not used for the binaries, in fact we should already have that for libcheck, and not require them for building the binaries. I believe it should be on the same repo, as they may need to get updated with code changes and that seems to be the common practice in other projects (e.g.: node projects have a `devDependencies` in `package.json`, rust projects have `dev-dependencies` in `Cargo.toml` and so on; also for example Redis repo has its tcl integration tests included).

Given this specific use case for partial requests, I don't believe any memcached client will expose an appropriated interface for the test we need, which is unfortunate, so I believe any scripting language will be good enough.

src/server/twemcache/data/process.c
```diff
@@ -198,21 +265,31 @@ _process_set(struct response *rsp, struct request *req)
...
+    status = _put(&istatus, req);
+    if (status == PUT_PARTIAL) {
+        return;
+    }
 
     INCR(process_metrics, add);
-    key = array_first(req->keys);
-    if (item_get(key) != NULL) {
+    it = (struct item *)req->reserved;
+    key = (struct bstring){it->klen, item_key(it)};
```
seppo0010: if `item_reserve` failed in `_put`, `req->reserved` will be null, and all this code will fail; check for `status != PUT_ERROR` earlier

src/server/twemcache/data/process.c
```diff
@@ -223,22 +300,30 @@ _process_add(struct response *rsp, struct request *req)
...
+        return;
+    }
 
     INCR(process_metrics, replace);
-    key = array_first(req->keys);
-    if (item_get(key) != NULL) {
-        item_delete(key);
-        status = item_insert(key, &(req->vstr), req->flag, time_reltime(req->expiry));
-        if (status == ITEM_OK) {
+    it = (struct item *)req->reserved;
```
seppo0010: same here ^^

src/server/twemcache/data/process.c
```diff
@@ -249,27 +334,37 @@ _process_replace(struct response *rsp, struct request *req)
...
+
+    status = _put(&istatus, req);
+    if (status == PUT_PARTIAL) {
+        return;
+    }
 
-    key = array_first(req->keys);
-    it = item_get(key);
-    if (it == NULL) {
+    it = (struct item *)req->reserved;
```
seppo0010: ^^

src/server/twemcache/data/process.c
```diff
@@ -306,8 +401,9 @@ _process_delta(struct response *rsp, struct item *it, struct request *req,
             status = item_update(it, &nval);
         } else {
             uint32_t dataflag = it->dataflag;
-            item_delete(key);
-            status = item_insert(key, &nval, dataflag, it->expire_at);
+            status = item_reserve(&it, key, &nval, nval.len, dataflag,
+                    it->expire_at);
```
seppo0010: check for status before inserting?

src/protocol/data/memcache/parse.c
```diff
@@ -273,29 +273,36 @@ _chase_uint(uint64_t *num, struct buf *buf, bool *end, uint64_t max)
...
+    bool complete = true;
 
     log_verb("parsing val (string) at %p", buf->rpos);
 
-    if (buf_rsize(buf) < vlen + CRLF_LEN) {
-        log_verb("buf %p has %"PRIu32" out of the %"PRIu32" bytes expected",
-                buf, buf_rsize(buf), vlen + CRLF_LEN);
-
-        return PARSE_EUNFIN;
+    if (rsize < nbyte + CRLF_LEN) {
```
thinkingfish: Got it. Yes! This is better :)

src/protocol/data/memcache/parse.c
```diff
@@ -782,31 +790,93 @@ _parse_req_hdr(struct request *req, struct buf *buf)
...
+    case REQ_PARSING: /* a new request */
+        log_verb("parsing buf %p into req %p", buf, req);
+        req->first = true;
+        status = _parse_req_hdr(req, buf);
+        if (status == PARSE_EUNFIN) {
+            log_verb("incomplete data: reset read position, jump back %zu bytes",
+                    buf->rpos - old_rpos);
+            /* return and start from beginning next time */
+            request_reset(req);
+            buf->rpos = old_rpos;
```
thinkingfish: You are right. I created a new test to capture this condition (and without the break it failed). Thanks for catching this!

src/server/twemcache/data/process.c
thinkingfish: will fix.

src/server/twemcache/data/process.c
thinkingfish: will fix.

src/server/twemcache/data/process.c
thinkingfish: will fix.

src/server/twemcache/data/process.c
thinkingfish: will fix.

src/server/twemcache/data/process.c
```diff
@@ -242,23 +245,25 @@ _process_set(struct response *rsp, struct request *req)
...
-        it = (struct item *)req->reserved;
-        key = (struct bstring){it->klen, item_key(it)};
-        item_insert(it, &key);
-        rsp->type = RSP_STORED;
-        INCR(process_metrics, set_stored);
-    } else { /* PUT_ERROR */
+    if (status == PUT_ERROR) {
         _error_rsp(rsp, istatus);
-        INCR(process_metrics, set_ex);
+        INCR(process_metrics, cas_ex);
```
seppo0010: `cas_ex`?

src/server/twemcache/data/process.c
```diff
@@ -242,23 +245,25 @@ _process_set(struct response *rsp, struct request *req)
     struct item *it;
     struct bstring key;
 
-    status = _put(&istatus, req);
+    status = _put(&istatus, req, true);
```
seppo0010: why are you deleting the existing item if `set` fails? If I understand this correctly, failing to set `key` to `value2` will delete existing `value1`, is that so?

thinkingfish: Correct. The reason has much to do with cache invalidation:

For cache, it's generally considered worse to have stale/incorrect data than no data at all. In other words, proactive cache invalidation is desirable for correctness, as long as that doesn't happen on a massive scale (e.g. entire cache invalidated at once). When someone is sending request to update a key, it usually implies that the existing value is no longer valid.

The only exception to this assumption is ```add```, which only wants to insert if the key does not already exist. ```add``` is most commonly used for immutable data in practice. Therefore, failing to write the new value is not ground for deleting the old value, which may very well be the same.

Let me add the reasoning as a note here.

thinkingfish: @seppo0010 I thought about it some more, and can see the appeal of not deleting upon exceptions. Basically it means a command that resulted in error has no side effects. (and client can retry).

The current behavior is inherited from [Twemcache](https://github.com/twitter/twemcache/blob/master/src/mc_ascii.c#L1136 ) (which in turn got it from memcached).

src/server/twemcache/data/process.c
thinkingfish: oops, copy paste error

test/protocol/data/memcache/check_memcache.c
```diff
@@ -310,6 +310,77 @@ START_TEST(test_set)
...
+#define SERIALIZED_2 "xyz"
+#define SERIALIZED_3 "\r\n"
+#define KEY "foo"
+#define VAL1 "XYZ"
+#define VAL2 "xyz"
+
+    int ret;
+    struct bstring key = str2bstr(KEY);
+    struct bstring val1 = str2bstr(VAL1);
+    struct bstring val2 = str2bstr(VAL2);
```
seppo0010: I prefer if `VAL1` and `VAL2` were not the same length as it can hide some arithmetic bug like subtracting remaining number of bytes instead of received bytes

test/protocol/data/memcache/check_memcache.c
```diff
@@ -630,6 +701,33 @@ START_TEST(test_decr_noreply)
...
+    /* parse (nothing should change) */
+    request_reset(req);
+    ret = parse_req(req, buf);
+    ck_assert_int_eq(ret, PARSE_EUNFIN);
+    ck_assert(req->rstate == REQ_PARSING);
+    ck_assert(buf->rpos == pos);
+#undef EXPIRY
+#undef FLAG
+#undef VAL
+#undef KEY
```
seppo0010: `EXPIRY`, `FLAG`, `VAL` and `KEY` were not defined, remove `undef`

test/protocol/data/memcache/check_memcache.c
thinkingfish: fixed

test/protocol/data/memcache/check_memcache.c
thinkingfish: fixed

thinkingfish:
Agree with you on having end-to-end functional tests. I am researching scripting options. Equivalent redis and memcached, twemcache tests all use an existing clients (redis with tcl + redis-cli, memcached with perl, twemcache with py-memcache). These are all reasonable choices. I prefer having something like an cli built into the repo at some point soon, so I want to think about an approach that's more similar to redis', but I'm not at all familiar with Tcl so I need to do some more research.

thinkingfish:
Starting a new issue to discuss end-to-end test strategy in #129 
I would like to keep it out of this PR, but I'm open to merge this change _after_ we implement/merge the integration tests.

Issue #129, opened 2017-02-17
end-to-end tests for pelikan servers


thinkingfish:
We need some end-to-end test cases and harness to make sure the servers behave as expected. Initially looking into the most basic, deterministic scenarios. In these cases a sequence of request/response pairs will be used, the requests are sent to the server and the responses have to match exactly.

Since we don't have an official client library in Pelikan just yet ( @kevyang wrote one for memcached, but it needs a bit of polishing which I will do soon), we can start with sending serialized strings and treating responses the same way. Under this approach, the only part in the response that needs attention is the cas value for ```gets``` commands. Exact values can be ensured if we locally track the number of writes sent to the server; alternatively we can use regex to parse out the responses.

I'm proposing separating the test cases from the harness to maximize readability and flexibility.

Test cases can be stored in a file with some delimiter between individual request/response pairs, e.g.:
```
>>>
set foo 0 0 3
bar
<<<
STORED

>>>
gets foo
<<<
VALUE foo 0 3 1
bar
END
```

And test harness can read from such a file, as well as a server config file to learn the whereabouts of the server, and play the tests against it. I've looked at test harness written in Perl, Tcl, Bash and Python, and I lean toward Python for readability, given integration tests don't have particular performance requirements.
Perl, Tcl and Python are pretty much universally installed on all POSIX compatible systems (not to mention bash), so availability is not a problem for any of them.
Anything else that's worth looking at, or is there an argument for/against any of them?

/cc @seppo0010

seppo0010:
👍

seppo0010:
I'm wondering is if such syntax may access the metrics? For example, the partial requests, since it's a feature that's mostly invisible for users, you'll probably want to assert partial requests were used.

I'm not sure if a domain specific language would be best, as you'll need to add a bunch of features for some cases, for example loops or somehow being able to create a large payload. Is it better than just using plain Python with a Client object?

You might also want to consider the possibility of having more than one concurrent client to test `cas`, although that may be out of the scope of this kind of tests.

thinkingfish:
Your points/concerns are totally valid. It is true a naive list of commands can make certain tests cumbersome. On the flip side, using a different client introduces other problems, for example, if I want to test a new command/data type, then I will need to implement the corresponding APIs in the dependent client, which may come from a 3rd party and now I need to start tracking upstream.

[...]

PR #130, opened 2017-02-21
Add basic python integration test

seppo0010:
kick off for issue #129 ; not complete, I want to gather feedback before moving on

### DIFF
--- a/test/integration/base.py
+++ b/test/integration/base.py
@@ -1,0 +1,113 @@
+import contextlib
+import errno
+import os
+import socket
+import subprocess
+import tempfile
+import unittest
+
+
+class PelikanTest(unittest.TestCase):
+    reserved_ports = set()
+
+    @staticmethod
+    def default_path():
+        return os.path.realpath(os.path.join(
+            os.path.dirname(os.path.abspath(__file__)),
+            '../../_build/_bin'
+        ))
+
+    # from http://stackoverflow.com/a/22106569
+    @staticmethod
+    def get_open_port(lowest_port = 10000, highest_port = None, bind_address = '', *socket_args, **socket_kwargs):
+        if highest_port is None:
+            highest_port = lowest_port + 100
+        while lowest_port < highest_port:
+            if lowest_port not in PelikanTest.reserved_ports:
+                try:
+                    with contextlib.closing(socket.socket(*socket_args, **socket_kwargs)) as my_socket:
+                        my_socket.bind((bind_address, lowest_port))
+                        this_port = my_socket.getsockname()[1]
+                        PelikanTest.reserved_ports.add(this_port)
+                        return this_port
+                except socket.error as error:
+                    if not error.errno == errno.EADDRINUSE:
+                        raise
+                    assert not lowest_port == 0
+                    PelikanTest.reserved_ports.add(lowest_port)
+            lowest_port += 1
+        raise Exception('Could not find open port')
+
+    def setUp(self):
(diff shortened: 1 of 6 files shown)


test/integration/base.py
```diff
@@ -0,0 +1,113 @@
...
+    reserved_ports = set()
+
+    @staticmethod
+    def default_path():
+        return os.path.realpath(os.path.join(
+            os.path.dirname(os.path.abspath(__file__)),
+            '../../_build/_bin'
+        ))
+
+    # from http://stackoverflow.com/a/22106569
```
thinkingfish: Can you split out the socket related logic into its own class and module/file? This will allow multiple test files to share the same logic instead of having all tests in a giant file.

test/integration/base.py
```diff
@@ -0,0 +1,113 @@
...
+            stderr=subprocess.PIPE,
+        )
+
+        # wait for the port to be listening; no great "ready" output present
+        # but it lists all configs, so this must exist
+        while 'name: server_port' not in self.server.stdout.readline():
+            pass
+
+        self.client = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        self.client.connect(('127.0.0.1', self.port))
```
thinkingfish: localhost instead of 127.0.0.1?

test/integration/base.py
```diff
@@ -0,0 +1,113 @@
...
+    @staticmethod
+    def default_path():
+        return os.path.realpath(os.path.join(
+            os.path.dirname(os.path.abspath(__file__)),
+            '../../_build/_bin'
+        ))
+
+    # from http://stackoverflow.com/a/22106569
+    @staticmethod
+    def get_open_port(lowest_port = 10000, highest_port = None, bind_address = '', *socket_args, **socket_kwargs):
```
thinkingfish: I'm not sure if we want this logic in the unit test:
1. for initial local use, assuming the default ports are available and throw an error when not seems OK;
2. if running in Twitter's container environment, the port range is specified by the container and unlikely to be the one(s) specified in the code here.

I think it might be easier to leave the server setup to a separate script. To summarize (including the comment on socket IO):
- a module that deals with all client-side network operations;
- a module that sets up the server, which takes executable location and port as input;
- an "orchestrator" that provides info to the server module and then passes the location of the server to the client module.

seppo0010:
@thinkingfish I addressed your feedback, I'm not sure if fully. Feel free to take over the PR if you want.

test/integration/pelikan_server.py
```diff
@@ -0,0 +1,39 @@
...
+            self.executable
+        )
+        self.server = subprocess.Popen((executable, self.config_file.name),
+            stdin=subprocess.PIPE,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+        )
+
+    def wait_ready(self):
+        # wait for the port to be listening; no great "ready" output present
```
thinkingfish: maybe use a simple command such as version on the admin port? But this is also fine for now.

thinkingfish:
@seppo0010 I may work on top of this PR later but it's in fine shape as a baseline. It's always good to have you on the team :)

test/integration/pelikan_server.py
seppo0010: The problem using a socket is that it fails to connect if you do it immediately and the server is not listening yet... reading from stdout is a lot more convenient!

Issue #39, opened 2015-11-13
Update admin protocol to take a second field (e.g. "stats slab")


kevyang:

thinkingfish:
#136 implements "stats slab"

PR #136, opened 2017-02-25
Add perslab metrics that help understand item size distribution

thinkingfish:
We deviate from memcached's slab metrics format because honestly it was not particular readable. Now they look like this:
```
stats slab
CLASS 1: chunk_size 48 item_keyval_byte 0 item_val_byte 0 item_curr 0 item_free 0 slab_curr 0
CLASS 2: chunk_size 56 item_keyval_byte 0 item_val_byte 0 item_curr 0 item_free 0 slab_curr 0
CLASS 3: chunk_size 64 item_keyval_byte 0 item_val_byte 0 item_curr 0 item_free 0 slab_curr 0
...
END
```

Compatibility is not important because internally we never exported these numbers anywhere, and the current table format is easier to read than one metric per line.

### DIFF
--- a/src/protocol/admin/parse.c
+++ b/src/protocol/admin/parse.c
@@ -71,19 +71,20 @@ admin_parse_req(struct request *req, struct buf *buf)
     }
     p = q = buf->rpos;
 
-    /* First find CRLF, this simplifies parsing. For admin port we don't care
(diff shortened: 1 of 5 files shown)


thinkingfish:
This addresses one of the main goals of #39

Issue #138, opened 2017-02-25
request_free gauge going negative


seppo0010:
While doing some debugging on travis problems I noticed `request_free` having a value of `-1` after doing a simple `get` to an empty key https://travis-ci.org/seppo0010/pelikan/jobs/205298464

thinkingfish:
looking

PR #140, opened 2017-02-25
fixing bug #138

thinkingfish:


thinkingfish:
@seppo0010 That's coming in my next PR which is integration test. I want to refactor the integration test before adding more tests.

PR #142, opened 2017-02-27
Test harness for #129

thinkingfish:


test/integration/base.py
```diff
@@ -1,50 +1,46 @@
-import unittest
-
-from .pelikan_client import PelikanClient
-from .pelikan_server import PelikanServer
-
-SERVER_PORT = 12345
-ADMIN_PORT = 12346
+from client import DataClient, AdminClient
```
seppo0010: I believe python's community prefers `.client` in this situation (and `client` will fail with python3) to prevent cases where names are ambiguous with top level modules.

src/server/twemcache/main.c
```diff
@@ -1,3 +1,4 @@
+#include "admin/process.h"
```
seppo0010: This looks out of place... why are you changing this file now, here, in this branch?

test/integration/base.py
```diff
@@ -1,50 +1,46 @@
...
+        """loading a test sequence from a file"""
+        self.seq = load_seq(fname)
+
+    def assertResponse(self, expected):
+        """receive and verify response (a list) matches expectation"""
+        if len(expected) > 0:
+            rsp = self.data_client.response()
+            self.assertEqual(len(rsp), len(expected))
+            for i in range(len(rsp)):
+                self.assertEqual(rsp[i], expected[i])
```
seppo0010: why not just
`self.assertEqual(rsp, expected)`?

test/integration/base.py
```diff
@@ -1,50 +1,46 @@
...
+
+    def assertStats(self, delta):
+        """delta, a dict, captures the expected change in a subset of metrics"""
+        stats = self.admin_client.stats()
+        for k in delta:
+            self.assertEqual(int(stats[k]) - int(self.stats[k]), delta[k])
```
seppo0010: This could use a third parameter (an error message) including `k`, otherwise getting `3 is not equal to 2` is hard to understand, specially since many parameters may be checked in the same stacktrace.

test/integration/loader.py
```diff
@@ -0,0 +1,58 @@
...
+    for m in metrics:
+        name,delta = m.strip().split(' ')
+        d[name.strip()] = int(delta)
+
+    return d
+
+
+def load_seq(fname):
+    """Load the test (a sequence of commands and asserts) from a file.
+    Each command contains one or more lines of request, leading with '>>> ', and
```
seppo0010: Empty line between title and body of the documentation text

test/integration/loader.py
```diff
@@ -0,0 +1,58 @@
...
+    one or more lines of response, leading with '<<< '. Commands are separated
+    by an empty line.
+    """
+    lines = open(fname).readlines()
+    if not re_empty.match(lines[-1]):  # ensure an empty line at the end
+      lines.append('\n')
+
+    seq = []
```
seppo0010: this structure is a bit more complicated than I expected... why don't you just make an array seq with three types of values, either request, response or stats and read it line by line? No need to aggregate them in a single sequence of all three... this is also prone to errors if you declare a req, then a stat, then a req, since both req will be executed before the stat (unless there's a blank line in between!).

test/integration/client.py
```diff
@@ -0,0 +1,119 @@
...
+        except Exception:
+            self.close()
+            raise
+
+
+class DataClient(TCPClient):
+    DELIM = '\r\n'
+
+    def __init__(self, *args, **kwargs):
+        super(DataClient, self).__init__(*args, **kwargs)
```
seppo0010: remove `__init__`

test/integration/client.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def request(self, req):
+        """send a (multi-line) request, req should be of a sequence type"""
+        data = DataClient.DELIM.join(req)
+        self.send(data + DataClient.DELIM)
```
seppo0010:
```python
for r in req:
    self.send(req + DataClient.DELIM)
```

less data copy, and looks less error prone

test/integration/client.py
```diff
@@ -0,0 +1,119 @@
...
+            raise Exception("response not terminated by " + DataClient.DELIM)
+        return rsp
+
+
+class AdminClient(TCPClient):
+    DELIM = '\r\n'
+    STATS_CMD = 'stats'
+
+    def __init__(self, *args, **kwargs):
+        super(AdminClient, self).__init__(*args, **kwargs)
```
seppo0010: remove

.travis.yml
```diff
@@ -93,4 +93,4 @@ script:
   - cmake ..
   - make -j
   - make check
-  - cd ../test && python -m unittest discover
+  - cd ../test/integration && python test_twemcache.py
```
seppo0010: why not just have a test_pelikan.py that searches for all sequences for twemcache, slimcache, etc.?

test/integration/base.py
thinkingfish: Ah I didn't know that- should have researched why . was prefixed.

Shall we request python3 in the script?

test/integration/base.py
thinkingfish: Because I'm not very good at Python or other modern high level language?

Will fix.

src/server/twemcache/main.c
thinkingfish: It was throwing a compilation warning at me when I was running the test. I could exclude this and put it on a separate PR.

.travis.yml
thinkingfish: Because test discovery right now is broken (due to how I wrote runTest method), so I need to fix it before use that style again.

test/integration/loader.py
thinkingfish: My goal was to enforce the test cases to be based on request-response pairs, so I actually would like to enforce the use of blank line between tests later, and enforce the ordering of request-response-stats. Line-by-line is simpler to write/load, but IMO not as pleasant to parse for a human reader.

test/integration/base.py
thinkingfish: Certainly. Will fix.

test/integration/loader.py
thinkingfish: added

test/integration/client.py
thinkingfish: True. Probably also good for testing partial requests.

test/integration/client.py
thinkingfish: removed, as well as in the adminclient

test/integration/loader.py
thinkingfish: You have a valid point about observing stats change mid-request though, which is something that we may want to do, and my assumption would force such sequence to be carried out by pretending we are sending two separate requests, which is not quite right.

I agree you suggestion is better in general, I would only add that blank lines can be added anytime to break up the flow visually, and can be ignored by the loader.

thinkingfish:
the server's ready() function does not work with python3 (works with python2), which I will also look into.

src/server/twemcache/main.c
seppo0010: It's fine, I just wanted to make sure it wasn't accidental

test/integration/base.py
```diff
@@ -29,14 +29,15 @@ def assertResponse(self, expected):
         if len(expected) > 0:
             rsp = self.data_client.response()
             self.assertEqual(len(rsp), len(expected))
```
seppo0010: remove

test/integration/base.py
seppo0010: did you change it?

test/integration/loader.py
seppo0010: It's unclear to me if you will make this change or keep it as is based on your comments.

thinkingfish: I tried the .client, and realized it needs to work with python3 as-is. I suggest we keep it as-is for now and create a new issue to investigate the best practice for this (maybe alongside unittest discovery)

test/integration/base.py
thinkingfish: fixed

thinkingfish:
Ooops sorry I forgot to merge in one of the fixed item. Will get that in with #143 PR.

PR #146, opened 2017-03-03
redis protocol: tokenizer, hash and zset

thinkingfish:
tokenizer deals with the basic serialization types. It currently handles all the elemental types but not nested arrays.

as far as parser is concerned the amount of work on top of tokenizer is minimum. The variable part of the argument list will be handled by processing logic.


src/protocol/data/redis/compose.h
```diff
@@ -0,0 +1,44 @@
...
+    COMPOSE_RSP_METRIC(METRIC_DECLARE)
+} compose_rsp_metrics_st;
+
+typedef enum compose_rstatus {
+    COMPOSE_OK          = 0,
+    COMPOSE_EUNFIN      = -1,
+    COMPOSE_ENOMEM      = -2,
+    COMPOSE_EINVALID    = -3,
+    COMPOSE_EOTHER      = -4,
+} compose_rstatus_t;
```
kevyang: According to the coding style guidelines, this should be compose_rstatus_i

src/protocol/data/redis/token.c
```diff
@@ -0,0 +1,339 @@
...
+    while (n > buf_wsize(*buf)) {
+        if (dbuf_double(buf) != CC_OK) {
+            log_debug("failed to write  %u bytes to buf %p: insufficient "
+                    "buffer space", n, *buf);
+
+            return COMPOSE_ENOMEM;
+        }
+    }
+
+    return CC_OK;
```
kevyang: COMPOSE_OK

src/protocol/data/redis/token.c
```diff
@@ -0,0 +1,339 @@
...
+{
+    parse_rstatus_t status;
+    int64_t len;
+
+    bstring_init(str);
+    status = _read_int(&len, buf, -1, BULK_MAXLEN);
+    if (status != PARSE_OK) {
+        return status;
+    }
+    if (len < 0) {
```
kevyang: what if len == 0?

src/protocol/data/redis/token.c
```diff
@@ -0,0 +1,339 @@
...
+    parse_rstatus_t status;
+
+    ASSERT(buf_rsize(buf) > 0);
+
+    log_verb("detecting the next element %p in buf %p", el, buf);
+
+    p = buf->rpos++;
+    switch (*p) {
+    case '+':
+        /* imple string */
```
kevyang: simple

src/protocol/data/redis/token.c
```diff
@@ -0,0 +1,339 @@
...
+}
+
+
+int
+compose_array_header(struct buf **buf, int nelem)
+{
+    struct buf *b;
+    size_t n = 1 + CRLF_LEN + CC_INT64_MAXLEN;
+
+    if (_check_buf_size(buf, n) != CC_OK) {
```
kevyang: COMPOSE_OK

src/protocol/data/redis/compose.h
thinkingfish: Yea you are right, and I noticed that too; let me create an issue to change this for all protocols.

src/protocol/data/redis/token.c
thinkingfish: fixed.

src/protocol/data/redis/token.c
thinkingfish: apparently that means an empty string, not a null/nil string. I should add a test.
So I think the following should be legal:
"$0\r\n\r\n"

thinkingfish: test added

src/protocol/data/redis/token.c
thinkingfish: fixed

src/protocol/data/redis/token.c
thinkingfish: fixed

src/protocol/data/redis/compose.h
thinkingfish: Created issue #149

src/protocol/data/redis/parse.h
```diff
@@ -0,0 +1,56 @@
...
+    PARSE_EOVERSIZE = -3,
+    PARSE_EINVALID  = -4,
+    PARSE_EOTHER    = -5,
+} parse_rstatus_t;
+
+void parse_setup(parse_req_metrics_st *req, parse_rsp_metrics_st *rsp);
+void parse_teardown(void);
+
+static inline bool
+key_valid(struct bstring *key)
```
seppo0010: unused?

src/protocol/data/redis/request.c
```diff
@@ -0,0 +1,213 @@
...
+    rstatus_i status;
+    struct request *req = cc_alloc(sizeof(struct request));
+
+    if (req == NULL) {
+        return NULL;
+    }
+
+    status = array_create(&req->token, ntoken, sizeof(struct element));
+    if (status != CC_OK) {
+        return NULL;
```
seppo0010: `cc_free(req);`? do we care?

src/protocol/data/redis/response.c
```diff
@@ -0,0 +1,205 @@
...
+    rstatus_i status;
+    struct response *rsp = cc_alloc(sizeof(struct response));
+
+    if (rsp == NULL) {
+        return NULL;
+    }
+
+    status = array_create(&rsp->token, ntoken, sizeof(struct element));
+    if (status != CC_OK) {
+        return NULL;
```
seppo0010: free rsp

src/protocol/data/redis/token.c
```diff
@@ -0,0 +1,339 @@
...
+
+    return PARSE_EUNFIN;
+}
+
+static inline int
+_write_int(struct buf *buf, int64_t val)
+{
+    size_t n = 0;
+
+    /* TODO: use cc_print_int64_unsafe when it's checked in */
```
seppo0010: address TODO?

src/protocol/data/redis/token.c
```diff
@@ -0,0 +1,339 @@
...
+/* this function does not handle array, which is a composite type */
+int
+compose_element(struct buf **buf, struct element *el)
+{
+    size_t n = 1 + CRLF_LEN;
+    struct buf *b;
+
+    ASSERT(el->type > 0);
+
+    /* first estimate size (overestimate for integers) */
```
seppo0010: overestimating for bulk as well

test/protocol/data/redis/check_redis.c
```diff
@@ -0,0 +1,540 @@
...
+ * test cases *
+ **************/
+
+/*
+ * token
+ */
+START_TEST(test_simple_string)
+{
+#define SERIALIZED "+foobar\r\n"
+#define STR "foobar"
```
seppo0010: how about this?
```c
#define STR "foobar"
#define SERIALIZED "+" STR "\r\n"
```

test/protocol/data/redis/check_redis.c
```diff
@@ -0,0 +1,540 @@
...
+
+    /* compose */
+    el.type = ELEM_STR;
+    el.bstr = str2bstr(STR);
+    ret = compose_element(&buf, &el);
+    ck_assert_msg(ret == len, "bytes expected: %d, returned: %d", len, ret);
+    ck_assert_int_eq(cc_bcmp(buf->rpos, SERIALIZED, ret), 0);
+
+    /* parse */
+    pos = buf->rpos + 1;
```
seppo0010: reusing `el` from the compose section makes this test less reliable as the production code may be not setting a property that is set by the previous portion of the test; break it up into two tests? use a second `el`-like variable instead? destroy its value in between?

test/protocol/data/redis/check_redis.c
seppo0010: (this comment also applies for many tests below)

test/protocol/data/redis/check_redis.c
seppo0010: (this comment also applies for many tests below)

test/protocol/data/redis/check_redis.c
```diff
@@ -0,0 +1,540 @@
...
+#define S_ECHO "*2\r\n$4\r\nping\r\n$5\r\nhello\r\n"
+#define S_ECHO2 "*3\r\n$4\r\nping\r\n$5\r\nhello\r\n$5\r\nworld\r\n"
+    int ret;
+    struct element *el;
+
+    test_reset();
+
+    /* simple ping */
+    buf_write(buf, S_PING, sizeof(S_PING) - 1);
+    ck_assert_int_eq(parse_req(req, buf), PARSE_OK);
```
seppo0010:
```c
    ck_assert_int_eq(req->type, REQ_PING);

```

test/protocol/data/redis/check_redis.c
```diff
@@ -0,0 +1,540 @@
...
+    ck_assert(el.bstr.len == sizeof(ERR) - 1);
+    ck_assert(el.bstr.data == pos);
+
+#undef ERR
+#undef SERIALIZED
+}
+END_TEST
+
+START_TEST(test_integer)
+{
```
seppo0010: test oversize?

test/protocol/data/redis/check_redis.c
```diff
@@ -0,0 +1,540 @@
...
+    ck_assert_int_eq(parse_req(req, buf), PARSE_OK);
+    ck_assert_int_eq(req->token->nelem, 3);
+#undef S_ECHO2
+#undef S_ECHO
+#undef ECHO
+#undef S_PING
+#undef QUIT
+}
+END_TEST
+
```
seppo0010: add test for incomplete bulk and another one for multi bulk (maybe even one case for each element type if you feel like it) and check that it returns EUNFIN and resets rpos

src/protocol/data/redis/compose.c
```diff
@@ -0,0 +1,92 @@
...
+#include "token.h"
+
+#include <cc_debug.h>
+#include <cc_print.h>
+
+#define COMPOSE_MODULE_NAME "protocol::redis::compose"
+
+static bool compose_init = false;
+static compose_req_metrics_st *compose_req_metrics = NULL;
+static compose_rsp_metrics_st *compose_rsp_metrics = NULL;
```
kevyang: there's two of these here

src/protocol/data/redis/response.c
```diff
@@ -0,0 +1,205 @@
+#include "response.h"
+
+#include "token.h"
+
+#include <cc_debug.h>
+#include <cc_mm.h>
+#include <cc_pool.h>
+
+#define RESPONSE_MODULE_NAME "protocol::memcache::response"
```
kevyang: protocol::redis::response

kevyang:
I guess I am kind of confused as to what exactly the purpose/scope of the token.[ch] module is. It seems to me that it can mostly be folded into parse and compose; why is it its own module?

deps/ccommon/src/cc_print.c
```diff
@@ -62,6 +81,27 @@ cc_print_uint64(char *buf, size_t size, uint64_t n)
...
+    }
+
+    if (n < 0) {
+        *buf++ = '-';
+        d--;
+    }
+
+    _print_uint64(buf, d, n);
+
+    return (n < 0) ? d + 1 : d;
```
thinkingfish: d is the number of bytes printed, so we need to return the _original_ d not the decremented value (which is then used to print the unsigned). But it is true it can be made a little simpler. I have posted the changes to ccommon.

src/protocol/data/redis/cmd_hash.h
```diff
@@ -0,0 +1,28 @@
...
+    ACTION( REQ_HLEN,           "hlen",         2  )\
+    ACTION( REQ_HMGET,          "hmget",        -3 )\
+    ACTION( REQ_HMSET,          "hmset",        -4 )\
+    ACTION( REQ_HSET,           "hset",         4  )\
+    ACTION( REQ_HSETNX,         "hsetnx",       4  )\
+    ACTION( REQ_HSTRLEN,        "hstrlen",      3  )\
+    ACTION( REQ_HVALS,          "hvals",        2  )\
+    ACTION( REQ_HSCAN,          "hscan",        -3 )
+
+/* "hlen KEY" == "*2\r\n$4\r\nhlen\r\n$3\r\nKEY\r\n" */
```
thinkingfish: Intend to add another parser function interpreting the CLI-friendly format, which is space delimitered, so this line makes sense as an exampleonce we have that.

src/protocol/data/redis/parse.c
```diff
@@ -0,0 +1,179 @@
...
+            request_reset(req);
+            buf->rpos = old_rpos;
+            return status;
+        }
+        nelem--;
+    }
+
+    status = _parse_cmd(req);
+    if (status != PARSE_OK) {
+        return status;
```
thinkingfish: fixed.

src/protocol/data/redis/parse.h
thinkingfish: This is intended for the functions processing actual commands, such as ```hmget KEY1 KEY2```. The processing logic should call this function to make sure the keys are valid.

Redis commands are structured such that it is hard to fully validate requests before actually processing them, since the meaning of each field is command-dependent and not self-described.

src/protocol/data/redis/request.c
thinkingfish: we do. Thanks for catching this.

src/protocol/data/redis/response.c
thinkingfish: fixed

src/protocol/data/redis/token.c
thinkingfish: done.

src/protocol/data/redis/token.c
thinkingfish: updated/clarified

test/protocol/data/redis/check_redis.c
thinkingfish: I thought about it before, it is less error-prone if we change test cases, on the other hand it is less readable by themselves. You are probably still right though it end up a little better.

test/protocol/data/redis/check_redis.c
thinkingfish: Yea we should use two variables that are separately reset at the beginning.

test/protocol/data/redis/check_redis.c
thinkingfish: Updated to use concatenation

thinkingfish: tests updated to adopt the style you recommended

test/protocol/data/redis/check_redis.c
thinkingfish: added

test/protocol/data/redis/check_redis.c
thinkingfish: separate variables used now

src/protocol/data/redis/response.c
thinkingfish: fixed

thinkingfish:
Current implementation of tokenizer is, and may better be named as, an implementation of RESP (REdis Serialization Protocol). The idea is that there are more than one way of serializing the same redis request (e.g. the "simple" serialization uses space as a delimiter and have everything on the same line ending with CRLF; Redis replication use cap'n'proto to serialize requests, etc.

Regardless of how a request is serialized, it should generate the same request object (same goes for responses). This is the reason behind having all the serialization/deserialization separated out from the rest of the parsing/composing.

test/protocol/data/redis/check_redis.c
thinkingfish: Good call, I caught two bugs!

test/protocol/data/redis/check_redis.c
```diff
@@ -0,0 +1,590 @@
...
+START_TEST(test_integer)
+{
+#define SERIALIZED_1 ":-1\r\n"
+#define SERIALIZED_2 ":9223372036854775807\r\n"
+#define SERIALIZED_3 ":128\r\n"
+#define INT_1 -1
+#define INT_2 9223372036854775807LL
+#define INT_3 128
+#define OVERSIZE ":19223372036854775807\r\n"
+#define INVALID ":123lOl456\r\n"
```
seppo0010: `#define INVALID_2 ":\r\n"`

test/protocol/data/redis/check_redis.c
```diff
@@ -0,0 +1,590 @@
...
+    ck_assert(el_p.num == INT_2);
+
+    buf_reset(buf);
+    el_c.num = INT_3;
+    ret = compose_element(&buf, &el_c);
+    ck_assert(ret == sizeof(SERIALIZED_3) - 1);
+    ck_assert_int_eq(cc_bcmp(buf->rpos, SERIALIZED_3, ret), 0);
+    ret = parse_element(&el_p, buf);
+    ck_assert_int_eq(ret, PARSE_OK);
+    ck_assert(el_p.num == INT_3);
```
seppo0010: missing `ck_assert(el_p.type == ELEM_INT);`

these three could use a local macro to avoid duplication?

```c
#define TEST_SERIALIZE(serialized, value)\
    buf_reset(buf);\
    el_c.num = value;\
    ret = compose_element(&buf, &el_c);\
    ck_assert(ret == sizeof(serialized) - 1);\
    ck_assert_int_eq(cc_bcmp(buf->rpos, serialized, ret), 0);\
    ret = parse_element(&el_p, buf);\
    ck_assert_int_eq(ret, PARSE_OK);\
    ck_assert(el_p.type == ELEM_INT);\
    ck_assert(el_p.num == value);
TEST_SERIALIZE(INT_1, SERIALIZED_1);
TEST_SERIALIZE(INT_2, SERIALIZED_2);
TEST_SERIALIZE(INT_3, SERIALIZED_3);
#undef TEST_SERIALIZE
```

src/protocol/data/redis/token.c
```diff
@@ -0,0 +1,337 @@
...
+            len++;
+            *num = *num * 10ULL + sign * (*buf->rpos - '0');
+        } else {
+            if (len > 0 && line_end(buf)) {
+                buf->rpos += CRLF_LEN;
+                log_vverb("parsed integer, value %"PRIi64, *num);
+
+                return PARSE_OK;
+            } else {
+                if (*buf->rpos == CR) {
```
thinkingfish: Yea it should. OK I think read_int needs to a bit further tuning, current logic flow is error-prone.

src/protocol/data/redis/token.c
```diff
@@ -0,0 +1,337 @@
...
+    if (_check_buf_size(buf, n) != CC_OK) {
+        return COMPOSE_ENOMEM;
+    }
+
+    b = *buf;
+    log_verb("write element %p in buf %p", el, b);
+
+    switch (el->type) {
+    case ELEM_STR:
+        n = buf_write(b, "+", 1) + _write_bstr(b, &el->bstr);
```
thinkingfish: Interesting.

test/protocol/data/redis/check_redis.c
thinkingfish: Well I think using an array is probably better, similar to test_unfin.

src/protocol/data/redis/token.c
```diff
@@ -0,0 +1,337 @@
...
+
+    case ELEM_NIL:
+        n += 2; /* "-1" */
+        break;
+
+    default:
+        return COMPOSE_EINVALID;
+    }
+
+    if (_check_buf_size(buf, n) != CC_OK) {
```
thinkingfish: fixed.

test/protocol/data/redis/check_redis.c
thinkingfish: added

test/protocol/data/redis/check_redis.c
thinkingfish: rewritten with array + loop

src/protocol/data/redis/token.c
thinkingfish: all statements split.

test/protocol/data/redis/check_redis.c
thinkingfish: fixed.

PR #148, opened 2017-03-16
[hotkey] add sampling into slimcache and twemcache

kevyang:


src/server/slimcache/data/process.c
```diff
@@ -69,6 +69,16 @@ _get_key(struct response *rsp, struct bstring *key)
             rsp->vstr = val.vstr;
         }
 
+        if (hotkey_enabled) {
```
thinkingfish: One thought: to decouple hotkey detection and (judiciously) breaking backward compatibility, can we first enable hotkey detection and log such occurrence to debug log ( or a separate log if you will), before trying to send the signal back to the client? Since at this point we haven't introduced the client changes yet, logging allows us to start observing hotkey without introducing or breaking client dependency.

kevyang: done.

PR #150, opened 2017-04-01
dump all keys; count size of all/subset of keys based on prefix

thinkingfish:
I got a request to find out how many keys there are in cache that belong to a particular dataset (represented by a specific prefix). I thought about it for a bit and realized this feature might be generally useful to perform a total key dump for analysis or get an dataset level overview of the cache, so I checked this code it.

Not sure if this is ready for merge at this point: big issue is thread-safety. Since we perform the scanning on the admin thread, to avoid reading dirty keys (keys that are being updated by the worker thread) requires some kind of locking/synchronization. This we can do rather cheaply by using an atomic instruction to let worker thread alter a flag when an hashbucket is being actively modified, and having the admin thread blocked if it runs into a dirty hashtable entry. But the one-off usage I'm asked to provide can be done on a staging host after live traffic is shut off, so it doesn't seem to warrant the extra rigor for now.

I just want to have a place where this feature is visible/checked in.

### DIFF
--- a/src/storage/slab/item.c
+++ b/src/storage/slab/item.c
@@ -1,7 +1,9 @@
 #include "slab.h"
 
 #include <cc_debug.h>
+#include <cc_log.h>
 
+#include <fcntl.h>
 #include <stdlib.h>
 #include <stdio.h>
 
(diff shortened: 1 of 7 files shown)


seppo0010:
re: race condition, could we do what Redis does for saving (a fork and copy-on-write)?

src/storage/slab/item.c
```diff
@@ -372,3 +374,76 @@ item_flush(void)
...
+            log_info("... %"PRIu32" out of %"PRIu32" buckets scanned ...", i,
+                    nbucket);
+        }
+    }
+
+    log_info("finish scanning all keys");
+}
+
+bool
+item_dump(void)
```
seppo0010: why not receive a prefix here as well?

thinkingfish:
First- let me reiterate that this PR is not ready for merge in its current form, nor is it intended (I only need a quick hack, since we will prevent the race by not sending traffic after a while).

I think a fork with copy-on-write is going to significantly slow down the main thread when write volume is high. COW has locking too, it's just not explicit from the application's perspective, and we can do better.

The best way I can think of is to use a flag per hash entry to mark a "dirty" entry. This can be done via atomic instructions which can be considered as a form of hardware "lock", but much faster than a futex or other software locks. The background thread will wait until the next hashtable entry becomes clean to proceed, and same for the worker thread. This works better because 1) HW locks are fast; 2) locking granularity is very fine (one per hashtable entry, which corresponds to, on average, less than 2 keys); 3) contention surface is very small (only around link/unlink, since the background thread doesn't try to update the hashtable).

src/storage/slab/item.c
thinkingfish: Because after I implemented the "dump" command they realized having a "count" is sufficient :smi

src/storage/slab/item.c
```diff
@@ -372,3 +374,76 @@ item_flush(void)
...
+        log_stderr("Could not create key dump - cannot open file");
+        return false;
+    }
+
+    for (uint32_t i = 0; i < nbucket; i++) {
+        struct item_slh *entry = &hash_table->table[i];
+        struct item *it;
+
+        SLIST_FOREACH(it, entry, i_sle) {
+            if (write(fd, item_key(it), it->klen) < it->klen) {
```
kevyang: would we also want to check for expiry?

thinkingfish: I guess there're reasons to do either/both.

The only case I can think of where someone wants expired keys to be dumped, is to investigate how much space in cache is occupied by already expired keys. But that is a legitimate question that we have been asked.

I think we can keep this change simple, while clarifying what the current logic does. In the future, if someone requests dumping only live objects, we can make it an option, e.g. "dump live" "dump expired" "dump all"

src/storage/slab/item.c
thinkingfish: Alright, now I'm resuming this PR for actual merge, I'll take your suggestion and bring them to feature parity.

PR #151, opened 2017-04-10
Bug fix: Some unfin conditions are not properly handled

thinkingfish:


thinkingfish:
Unit tests added for both unfinished requests and additional unfin token (array "header")

PR #156, opened 2017-07-04
Pubsub server

thinkingfish:
This is a pubsub implementation that's compatible with Redis' pub/sub, but it's a standalone server, not mixed with other redis features.

Also, the publisher does not need to wait for the message fanout to complete before receiving a response, which fully decouples publishing from subscribing.

This is a prototype and many error/corner cases are not yet handled. Nor are unit tests added for the corresponding modules.


src/storage/pubsub/listener.c
```diff
@@ -0,0 +1,207 @@
...
+        prev = l, l = SLIST_NEXT(l, l_sle)) {
+        if (l->s == s) {
+            break;
+        }
+    }
+
+    if (prev == NULL) {
+        SLIST_REMOVE_HEAD(bucket, l_sle);
+    } else {
+        SLIST_REMOVE_AFTER(prev, l_sle);
```
seppo0010: what would happen if `s` was not on the list? `prev`'d be the last element and it'll segfault or something? should there be an `ASSERT(l != NULL)`?

src/storage/pubsub/topic.c
```diff
@@ -0,0 +1,215 @@
...
+    if (t->idx == NULL) {
+        cc_free(t);
+        return NULL;
+    }
+
+    topic_reset(t);
+    t->name.len = name->len;
+    t->name.data = cc_alloc(t->name.len);
+    if (t->name.data == NULL) {
+        cc_free(t);
```
seppo0010: missing `cc_free(t->idx);`

src/storage/pubsub/topic.h
```diff
@@ -0,0 +1,33 @@
+#pragma once
+
+#include "index.h"
+
+#include <cc_bstring.h>
+
+/*
+ * a topic is an endpoint that clients can subscribe to, equivalent to
+ * "channel" in the original redis protocol.
```
seppo0010: is there any reason to call it topic instead of channel? do we refer to something else as channel? you could add the reason to the comment here

src/server/pubsub/admin/process.c
```diff
@@ -0,0 +1,68 @@
...
+}
+
+void
+admin_process_teardown(void)
+{
+    log_info("tear down the %s module", PUBSUB_ADMIN_MODULE_NAME);
+    if (!admin_init) {
+        log_warn("%s has never been setup", PUBSUB_ADMIN_MODULE_NAME);
+    }
+
```
seppo0010: `cc_free(buf);`?

src/storage/pubsub/listener.h
```diff
@@ -0,0 +1,32 @@
+#pragma once
+
+#include "index.h"
+
+#include <stream/cc_sockio.h>
+
+/*
+ * a listener is a client that has subscribed to at least one channel
```
seppo0010: I thought that a listener may be subscribed to 0 channels (after unsubscribe it is not deleted)?

thinkingfish: The connection is still around, but no longer a listener. This is explained on redis.io in the [pubsub section](https://redis.io/topics/pubsub):

> unsubscribe: means that we successfully unsubscribed from the channel given as second element in the reply. The third argument represents the number of channels we are currently subscribed to. When the last argument is zero, we are no longer subscribed to any channel, and the client can issue any kind of Redis command as we are outside the Pub/Sub state.

src/storage/pubsub/topic.h
thinkingfish: Oh yea I should explain. mostly ccommon (which Pelikan depends on and includes) already have the concept of a "channel" to mean some kind of connection (tcp, uds, pipe...)

src/core/data/pubsub.c
```diff
@@ -0,0 +1,302 @@
...
+
+        event_del(ctx->evb, hdl->wid(c));
+        event_add_write(ctx->evb, hdl->wid(c), s);
+    } else if (status == CC_ERROR) {
+        c->state = CHANNEL_TERM;
+    }
+
+    if (processor->write(s) < 0) {
+        log_debug("handler signals channel termination");
+        s->ch->state = CHANNEL_TERM;
```
thinkingfish: done

src/server/pubsub/data/process.c
```diff
@@ -0,0 +1,337 @@
...
+ * reply: # listeners subscribed
+ */
+static void
+command_publish(struct response *rsp, struct request *req, struct buf_sock *s)
+{
+    struct element *el_t, *el_m; /* topic & message */
+    struct element el_r = {.type = ELEM_INT}; /* reply, an integer */
+    struct listener *l;
+    struct topic *t;
+    uint32_t nsub = 0;
```
thinkingfish: you are right, I changed it.

src/server/pubsub/data/process.c
```diff
@@ -0,0 +1,337 @@
...
+            /* handle error */
+            topic_put(t);
+        }
+
+        log_verb("subscribing to topic %.*s", el->bstr.len, el->bstr.data);
+        if (!topic_add_listener(t, l)) {
+            log_debug("listener not added");
+        }
+        if (!listener_add_topic(l, t)) {
+            log_debug("topic not added");
```
thinkingfish: Let me keep you comment in mind for the moment- I need to handle errors in a number of other places as well but it's not immediately clear what Redis does/should do in this case. I'll research more and come back to this.

src/storage/pubsub/listener.c
thinkingfish: You are right, the way I'm using it now assumes `l != NULL`, but I might as well allow passing in an invalid listener to simplify the deletion logic.

Eventually I want to add pooling to listener (and topic) to avoid calling malloc & free all the time, I will later think about if I want to allow people to create/destroy listener objects directly.

src/storage/pubsub/topic.c
thinkingfish: fixed

src/server/pubsub/admin/process.c
thinkingfish: fixed

src/core/core.c
```diff
@@ -10,7 +10,7 @@
 #include <sysexits.h>
 
 void
-core_run(void *arg_worker)
+worker_run(void *arg_worker)
```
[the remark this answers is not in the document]
kevyang: maybe rename these to core_worker_run and core_pubsub_run

PR #161, opened 2018-05-03
Handle empty buffers in resp parsing correctly, return false for array check and EUNFIN for general parsing

thinkingfish:


src/protocol/data/redis/token.c
```diff
@@ -186,7 +186,7 @@ token_is_array(struct buf *buf)
 {
     ASSERT(buf != NULL);
 
-    return *(buf->rpos) == '*';
+    return (buf_rsize(buf) > 0 && *(buf->rpos) == '*');
```
seppo0010: It makes more sense to return `parse_rstatus_t` here since both callers are validating the size of the read buffer...

how about something like this?

```c
parse_rstatus_t
token_is_array(struct buf *buf)
{
    ASSERT(buf != NULL)
    if (buf_rsize(buf) == 0) {
      return PARSE_EUNFIN;
    }
    if (*(buf->rpos) == '*') {
      return PARSE_OK;
    }
    return PARSE_EINVALID;
}
```

And callers should

```c
// ...
parse_rstatus_t status = token_is_array(buf);
if (status != PARSE_OK) {
    return status;
}
// ...
```

thinkingfish: I don't think using parse_rstatus_t as the return type really makes sense for this function. 

And exactly because both callers are validating the size of the read buffer, they wouldn't proceed assuming they have a primitive element at hand when this function returns false, but instead would always check if they have an element at all.

thinkingfish: This is one of those functions that would be declared as a protected member if we were using an OO language. The only module that should use this one directly is the parser.

PR #164, opened 2018-05-17
add subsecond timer

kevyang:


thinkingfish:
Do you think it makes sense to convert this module to use the timer in ccommon? It tracks internal duration in ns and provides API to convert it to any unit, so we can export now_* at all the common granularities.

kevyang:
yes. I can work on this change.

PR #165, opened 2018-05-18
add remove_val API implementation, allow found to only return state

thinkingfish:
the remove val API makes it easier to implement list command such as LREM.

allow find to accept `NULL` for both `ze` and `idx` so we can call this function just to check if a value exists or not, which is useful in testing and some higher level APIs.

### DIFF
--- a/src/data_structure/ziplist/ziplist.c
+++ b/src/data_structure/ziplist/ziplist.c
@@ -338,7 +338,7 @@ ziplist_rstatus_e
 ziplist_remove_val(uint32_t *removed, ziplist_p zl, const struct blob *val,
         int64_t count)
 {
-    uint32_t nentry, len, atmost;
+    uint32_t len, atmost;
     int64_t i = 0;
     zipentry_p z;
     uint8_t *end;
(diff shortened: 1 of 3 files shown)


src/data_structure/ziplist/ziplist.c
```diff
@@ -328,6 +325,71 @@ ziplist_find(zipentry_p *ze, int64_t *idx, const ziplist_p zl, const struct blob
...
+    if (zl == NULL || val == NULL) {
+        return ZIPLIST_ERROR;
+    }
+
+    if (val->type == BLOB_TYPE_UNKNOWN || val->type >= BLOB_TYPE_SENTINEL ||
+            (val->type == BLOB_TYPE_STR && val->vstr.len > ZE_STR_MAXLEN)) {
+        return ZIPLIST_EINVALID;
+    }
+
+    /* count == 0 means remove all, get the absolute value of count otherwise */
```
kevyang: Do you think we can find an alternative value to indicate remove all? I think having 0 be this value can be quite misleading, and somebody who has not read this comment may just assume a call to remove with count == 0 will just be a no op.

kevyang: it should work without needing to process count at all with int64 max and min, maybe we can define some macros to indicate these values

thinkingfish: the `count == 0` behavior is borrowed from `LREM` syntax of Redis (https://redis.io/commands/lrem), I don't feel strongly about this either way but am happy to look for alternatives.

For example, we can perform the same check as in remove where count == 0 is an error, and when we implement the redis command such as `LREM` we can translate the 0 there to `-UINT32_MAX` ?

kevyang: I see, I'm fine either way as well, especially if ziplist is only going to be used for redis protocol for the foreseeable future. I'll defer to your judgment here.

PR #176, opened 2018-06-30
CDB front and backend 

slyphon:
This PR contains an implementation of a CDB reader in Rust that builds and is linked with pelikan to create a memcache-protocol based frontend with a CDB backend. This change includes cmake/cargo integration adapted from [CMakeRust](https://github.com/wayk/CMakeRust).

The implementation adds an embedded buffer to the memcache response struct that is reused on each request/response cycle. This buffer is linked to the value `bstring` struct, passed into the CDB reader, and therefore we are able to do a single copy into the buffer that will be used with the write call on the socket.


src/server/cdb/main.c
```diff
@@ -0,0 +1,235 @@
...
+    debug_teardown();
+    log_teardown();
+}
+
+static struct CDBHandle*
+setup_cdb_handle(void)
+{
+    cdb_setup();
+
+    char *cdb_file_path = strdup("/Users/jsimms/pelikan/dict.cdb");
```
slyphon: herp derp

config/cdb.conf
```diff
@@ -0,0 +1,17 @@
...
+debug_log_file: cdb.log
+debug_log_nbuf: 1048576
+
+klog_file: cdb.cmd
+klog_backup: cdb.cmd.old
+klog_sample: 100
+klog_max: 1073741824
+
+# TODO: this is _obviously_ temporary
+# cdb_file_path: /Users/jsimms/git/tub/cdb.rs/dict.cdb
```
slyphon: I wasn't sure how the path to the cdb file should be specified, i.e. config file or command-line arg

src/protocol/data/memcache/response.c
```diff
@@ -18,6 +18,8 @@ struct bstring rsp_strings[] = {
 FREEPOOL(rsp_pool, rspq, response);
 static struct rsp_pool rspp;
 static bool rspp_init = false;
+/* TODO(simms): this should probably be readable from rust */
+static size_t rsp_val_buf_size = RSP_VAL_BUF_SIZE;
```
slyphon: I wasn't sure how to make this visible in rust. I guess having a function rust can call to get the value would be one way.

src/server/cdb/data/process.c
```diff
@@ -0,0 +1,500 @@
...
+
+static bool
+_get_key(struct response *rsp, struct bstring *key)
+{
+    /* this is a slight abuse of the bstring API. we're setting
+     * the data pointer to point to the vbuf that was allocated on the struct
+     * and we're setting the len of the buffer to the allocation size. This is
+     * so that we can create a rust slice from this information. */
+    rsp->vstr.data = rsp->vbuf;
+    rsp->vstr.len = RSP_VAL_BUF_SIZE;
```
slyphon: This is where the newly added `rsp->vbuf` gets hooked up for use on the rust side.

src/server/cdb/data/process.c
```diff
@@ -0,0 +1,500 @@
...
+
+    struct bstring *vstr = cdb_get(cdb_handle, key, &(rsp->vstr));
+
+    if (vstr != NULL) {
+        rsp->type = RSP_VALUE;
+        rsp->key = *key;
+        rsp->flag = 0;
+        rsp->vcas = 0;
```
slyphon: I was operating under the assumption that `rsp->flag` and `rsp-vcas` were ok to leave at 0 here for a read-only backing store

src/storage/cdb/cdb_ccommon/build.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    let mut config_h_dir = cbd.clone();
+    config_h_dir.push("ccommon");
+
+    let mut lib_dir = cbd.clone();
+    lib_dir.push("lib");
+
+    println!("cargo:rustc-link-search-native={}", lib_dir.to_str().unwrap());
+
+    // XXX: assume _build dir at pelikan top level
```
slyphon: TODO: remove, cruft

src/storage/cdb/cdb_ccommon/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+pub mod bindings {
+    #![allow(non_upper_case_globals)]
+    #![allow(non_camel_case_types)]
+    #![allow(non_snake_case)]
+    #![allow(dead_code)]
+    include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
+}
```
slyphon: This block, cargo-culted from the bindgen user guide, pulls all of the generated types in to the 'bindings' mod

src/protocol/data/memcache/response.c
slyphon: in a follow-up branch i've implemented this as a param that gets passed through to the cdb handle creation method.

src/server/cdb/data/process.c
thinkingfish: yep, totally fine.

src/server/cdb/data/process.c
```diff
@@ -0,0 +1,493 @@
...
+        } else {
+            INCR(process_metrics, get_key_miss);
+        }
+    }
+    r->type = RSP_END;
+
+    log_verb("get req %p processed, %d out of %d keys found", req, req->nfound, i);
+}
+
+static void
```
thinkingfish: For RO backend this would be of no use. Delete?

src/server/cdb/data/process.c
```diff
@@ -0,0 +1,493 @@
...
+    switch (req->type) {
+    case REQ_GET:
+        _process_get(rsp, req);
+        break;
+
+    case REQ_GETS:
+        _process_gets(rsp, req);
+        break;
+
+    case REQ_DELETE:
```
thinkingfish: I think you may save some space by collapsing these cases and apply something like `_process_unsupported()` to all of them.

src/server/cdb/admin/process.c
```diff
@@ -0,0 +1,98 @@
...
+    log_info("Mr. Gorbechev, tear down this module [%s]", CDB_ADMIN_MODULE_NAME);
+    if (!admin_init) {
+        log_warn("%s has never been setup", CDB_ADMIN_MODULE_NAME);
+    }
+
+    admin_init = false;
+}
+
+static void
+_admin_stats_slab(struct response *rsp, struct request *req)
```
thinkingfish: remove this one?

src/server/cdb/admin/process.c
```diff
@@ -0,0 +1,98 @@
...
+static void
+_admin_stats_default(struct response *rsp, struct request *req)
+{
+    procinfo_update();
+    rsp->data.data = buf;
+    rsp->data.len = print_stats(buf, cap, (struct metric *)&stats, nmetric);
+}
+
+static void
+_admin_stats(struct response *rsp, struct request *req)
```
thinkingfish: merge the default function into this one

src/protocol/data/memcache/response.c
```diff
@@ -52,6 +54,13 @@ response_create(void)
         return NULL;
     }
 
+    rsp->vbuf = cc_alloc(rsp_val_buf_size);
```
thinkingfish: RSP_VAL_BUF_SIZE is relatively large. If we allocate one such buffer per response (which I've been assigning/caching one per connection), you may end up using a lot of memory when the connection count is high. E.g. my current "high connection count" test has 10K+ connections so that would mean 10GB buffer being allocated.

In the single-threaded context, and size-adjustable `wbuf` for each stream (currently implemented using doubling buffer), I think it's possible to only allocate this buffer once, i.e. per-thread, and copying the content into the connection's `wbuf` before we turn to the next request or connection.

To merge this PR into master, I think it would be preferably that we don't affect the resource footprint of the other binaries dependent on `memcache response` objects, although, since `pelikan_twemcache` etc don't ever access this field, I don't think they will affect the RSS, so at least the current test jobs won't be OOM-killed.

src/server/cdb/data/process.c
```diff
@@ -0,0 +1,368 @@
...
+
+#define OVERSIZE_ERR_MSG    "oversized value, cannot be stored"
+#define DELTA_ERR_MSG       "value is not a number"
+#define OOM_ERR_MSG         "server is out of memory"
+#define CMD_ERR_MSG         "command not supported"
+#define OTHER_ERR_MSG       "unknown server error"
+
+static bool process_init = false;
+static process_metrics_st *process_metrics = NULL;
+static bool allow_flush = ALLOW_FLUSH;
```
thinkingfish: CDB should have no use of the flush option

src/server/cdb/data/process.c
```diff
@@ -0,0 +1,368 @@
...
+{
+    log_verb("processing req %p, write rsp to %p", req, rsp);
+    INCR(process_metrics, process_req);
+
+    switch (req->type) {
+    case REQ_GET:
+        _process_get(rsp, req);
+        break;
+
+    case REQ_GETS:
```
thinkingfish: I think this can also be moved to the invalid section, `gets` is used for  subsequent`cas`

slyphon:
should be good-to-go now

PR #178, opened 2018-07-02
Rewrite2

thinkingfish:


CMakeLists.txt
```diff
@@ -39,9 +39,11 @@ option(TARGET_REDIS "build redis binary" ON)
 option(TARGET_SLIMREDIS "build slimredis binary" ON)
 option(TARGET_SLIMCACHE "build slimcache binary" ON)
 option(TARGET_TWEMCACHE "build twemcache binary" ON)
-
+option(TARGET_CDB "build cdb binary" ON)
```
thinkingfish: default OFF

CMakeLists.txt
```diff
@@ -138,5 +144,14 @@ message(STATUS "HAVE_SIGNAME: " ${HAVE_SIGNAME})
 message(STATUS "HAVE_BACKTRACE: " ${HAVE_BACKTRACE})
 message(STATUS "HAVE_BIG_ENDIAN: " ${HAVE_BIG_ENDIAN})
 
+if(DUMP_ALL)
```
thinkingfish: Nice.

cmake/CMakeDetermineRustCompiler.cmake
```diff
@@ -0,0 +1,25 @@
+
+if(NOT CMAKE_Rust_COMPILER)
```
thinkingfish: Ideally I think it'd be nice to turn on cdb compiling if Rust is detected on the host, otherwise the target is skipped. Similar how test target is treated depending on whether check can be found or not.

src/protocol/data/memcache/response.h
```diff
@@ -80,6 +82,10 @@ struct response {
 
     struct bstring          key;        /* key string */
     struct bstring          vstr;       /* value string */
+    char                    *vbuf;      /* vbuf is a buffer of RSP_VAL_BUF_SIZE that processors can use by
```
thinkingfish: I think in a stable version this needs to be a server decision. So the capacity of the buffer will be announced to the storage module and storage can return the size of the data while copying things into this buffer. The treatment of return value in `snprintf` can be borrowed- a greater than capacity value means copy is incomplete. 

`vstr` above can be reused perhaps? `vstr.len` can carry the capacity of the buffer when passed in, `vstr.data` points to the buffer; and upon return, `vstr.len` can be updated to mean the number of bytes intended. The caller should have a copy of the original value of the buffer size so it can compare the return with that copy to see if further read is necessary to complete the response.

src/server/cdb/admin/process.c
```diff
@@ -0,0 +1,113 @@
...
+{
+    log_info("tear down the %s module", CDB_ADMIN_MODULE_NAME);
+    if (!admin_init) {
+        log_warn("%s has never been setup", CDB_ADMIN_MODULE_NAME);
+    }
+
+    admin_init = false;
+}
+
+static void
```
thinkingfish: can be deleted

src/server/cdb/data/process.h
```diff
@@ -0,0 +1,81 @@
...
+    ACTION( get_key,           METRIC_COUNTER, "# keys by get"         )\
+    ACTION( get_key_hit,       METRIC_COUNTER, "# key hits by get"     )\
+    ACTION( get_key_miss,      METRIC_COUNTER, "# key misses by get"   )\
+    ACTION( get_ex,            METRIC_COUNTER, "# get errors"          )\
+    ACTION( gets,              METRIC_COUNTER, "# gets requests"       )\
+    ACTION( gets_key,          METRIC_COUNTER, "# keys by gets"        )\
+    ACTION( gets_key_hit,      METRIC_COUNTER, "# key hits by gets"    )\
+    ACTION( gets_key_miss,     METRIC_COUNTER, "# key misses by gets"  )\
+    ACTION( gets_ex,           METRIC_COUNTER, "# gets errors"         )\
+    ACTION( delete,            METRIC_COUNTER, "# delete requests"     )\
```
thinkingfish: most of these can be deleted? I think the generic client error is sufficient.

thinkingfish:
Closing this one as a newer branch is in the PR list.

Issue #169, opened 2018-05-28
Race condition on line 22 of src/time/time.c


geeknik:
I compiled pelikan with `clang-7` and `-fsanitize=thread` without error. I then ran ./pelikan_pingserver and was greeted with the following:

```
launching server with default values.
set up the ccommon::log module
Set up the ccommon::debug module
create logger with filename (null) cap 0
[Mon May 28 17:34:41 2018][INFO] /root/pelikan/deps/ccommon/src/cc_signal.c:77 override handler for SIGSEGV

**SNIP**

[Mon May 28 17:34:41 2018][INFO] /root/pelikan/deps/ccommon/src/event/cc_epoll.c:90 epoll fd 9 with nevent 1024
name: daemonize                       type: boolean          current: no                   ( default: no                   )

**SNIP**

name: tcp_poolsize                    type: unsigned int     current: 0                    ( default: 0                    )
==================
WARNING: ThreadSanitizer: data race (pid=31778)
  Write of size 4 at 0x0000011318a8 by thread T2:
    #0 time_update /root/pelikan/src/time/time.c:22:9 (pelikan_pingserver+0x4b7786)
    #1 _server_evwait /root/pelikan/src/core/data/server.c:301:5 (pelikan_pingserver+0x4b4a34)
    #2 core_server_evloop /root/pelikan/src/core/data/server.c:310 (pelikan_pingserver+0x4b4a34)

  Previous write of size 4 at 0x0000011318a8 by thread T1:
    #0 time_update /root/pelikan/src/time/time.c:22:9 (pelikan_pingserver+0x4b7786)
    #1 _worker_evwait /root/pelikan/src/core/data/worker.c:286:5 (pelikan_pingserver+0x4b58e4)
    #2 core_worker_evloop /root/pelikan/src/core/data/worker.c:297 (pelikan_pingserver+0x4b58e4)

  Location is global 'now' of size 4 at 0x0000011318a8 (pelikan_pingserver+0x0000011318a8)

  Thread T2 (tid=31781, running) created by main thread at:
    #0 pthread_create /b/build/slave/linux_upload_clang/build/src/third_party/llvm/compiler-rt/lib/tsan/rtl/tsan_interceptors.cc:965:3 (pelikan_pingserver+0x449da5)
    #1 core_run /root/pelikan/src/core/core.c:29:11 (pelikan_pingserver+0x4b5c93)
[...]

PR #187, opened 2018-07-09
revamp timer

kevyang:
- add subsecond timer utility
- option to specify absolute, relative, or memcache style timestamps, with matching reltime function
- remove race condition with server/worker threads both updating time (#169) 
- unit tests


src/time/time.h
```diff
@@ -1,95 +1,118 @@
...
+/*
+ * Get time relative to process start, where t is number of seconds from now.
+ */
+static inline rel_time_t
+time_reltime(uint32_t t)
+{
+    return (rel_time_t)t + time_now();
+}
+#elif defined TIME_MEMCACHE
+#define TIME_MAXDELTA (time_t)(60 * 60 * 24 * 30)
```
thinkingfish: Can you move and preserve the comment quoting the memcached protocol's treatment of expiry around here?

src/time/time.h
```diff
@@ -1,95 +1,114 @@
...
 static inline time_t
 time_started(void)
 {
-    return time_start;
+    return __atomic_load_n(&time_start, __ATOMIC_RELAXED);
+}
+
+/* Get the current time in seconds since the process started */
+static inline rel_time_t
+time_now(void)
```
thinkingfish: Do you think we can rename these primitives to make it more obvious when the clock started?
e.g., we can have:
`process_time_sec()`, `process_time_ms()` .... that were started when the process started.

src/time/time.h
```diff
@@ -1,95 +1,114 @@
...
+ * timestamp of when the server came up.
+ *
+ * In relative timestamp only mode, t is treated as a time relative to time now,
+ * and is converted into time since process start.
+ *
+ * In memcached compatibility mode, t is treated as it is in memcached, that is,
+ * if it is greater than 30 days, it is treated as an absolute timestamp;
+ * otherwise, it is treated as a relative timestamp.
+ */
+#if defined TIME_ABSOLUTE
```
thinkingfish: I'm generally in favor of command line option (over compile switch), because then we can use the same binary for different settings without recompiling.

Also, the naming here is backward-compatible but may be challenged given the expansion to include finer clocks.
Essentially, we want `time_reltime()` to give us a timestamp that has 0 set to when the process started, which is used to determine eviction/expiring. Here we can match the time_now signatures to indicate different granularity, e.g. `time_unix2process_sec(uint32_t t)`, `time_unix2process_ms(uint64_t t)`...

src/time/time.h
```diff
@@ -1,95 +1,114 @@
...
- *
- * Memcached server uses a timer with second granularity, which starts as the
- * process starts, and is set to 2 initially to avoid having 0 aged items.
- *
- * On systems where size(time_t) > sizeof(uint32_t), this gives
- * us space savings over tracking absolute unix time of type time_t
+ * For less granular time, rel_time_t will suffice and gives space savings.
+ * For more granular time, rel_time_fine_t gives additional precision.
  */
 typedef uint32_t rel_time_t;
```
thinkingfish: I think the original typedef wasn't necessarily the best choice of name. Is this relative because the process itself is started at arbitrary time measured by Unix time, or is this relative because we are comparing against the process's clock? I would like to use this as an opportunity to improve that naming, e.g., how about `typedef uint32_t process_time_u` (also fixing the naming convention according to our style guide to move away from `_t` in applications). So now we know we are talking about time whose zero point is set to the time the process started.

src/time/time.h
```diff
@@ -1,95 +1,114 @@
...
+    return (rel_time_t)(t - time_started());
 }
-
-/* Get time relative to process start given absolute time */
+#elif defined TIME_RELATIVE
+/*
+ * Get time relative to process start, where t is number of seconds from now.
+ */
+static inline rel_time_t
+time_reltime(uint32_t t)
```
thinkingfish: Here we maybe able to name it as `time_delta2process_sec()`, etc

src/time/time.h
```diff
@@ -1,95 +1,114 @@
...
+}
+#elif defined TIME_MEMCACHE
+#define TIME_MAXDELTA (time_t)(60 * 60 * 24 * 30)
+/*
+ * Get time relative to process start given absolute time, memcached style.
+ * If a timestamp is greater than 30 days, then it is treated as an absolute,
+ * unix timestamp. Otherwise, it is treated as relative to now.
+ */
 static inline rel_time_t
 time_reltime(uint32_t t)
```
thinkingfish: Here we can call it `time_memcache2process_sec()`

For all the above, if the name seems way too long, we can use something else, or shorten them to be `time_u2p_*`, `time_d2p_*` and `time_m2p_sec()` (no need for finer-grain timestamp for memcache)

src/time/time.h
```diff
@@ -1,95 +1,114 @@
...
+/* Get the current time in nanoseconds since the process started */
+static inline rel_time_fine_t
+time_now_ns(void)
+{
+    return __atomic_load_n(&now_ns, __ATOMIC_RELAXED);
+}
+
+/* Get the current absolute time (not since process began) */
 static inline time_t
 time_now_abs(void)
```
thinkingfish: This function probably could be called `time_unix_sec()` and become more apparent.

CMakeLists.txt
```diff
@@ -7,8 +7,10 @@ enable_testing()
 macro(set_platform system_name)
     if(${system_name} MATCHES "Darwin")
         set(OS_PLATFORM "OS_DARWIN")
+        add_definitions(-DOS_DARWIN)
```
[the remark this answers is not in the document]
kevyang: `struct duration` can't be declared in pelikan without these definitions, since the definition of them in ccommon relies on these definitions being there. Without this, `struct duration` in cc_timer.h is just a forward declaration.

src/time/time.h
thinkingfish: Actually, the real problem with compile-time option is that if Redis wants one style of timestamp but twemcache wants another it would be really hard to compile them at once.

src/util/define.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+#include <config.h>
+
+#if defined HAVE_TIME_RELATIVE
```
thinkingfish: these can be removed I think?

src/server/slimredis/data/cmd_bitmap.c
```diff
@@ -75,7 +76,9 @@ _add_key(struct response *rsp, struct bstring *key)
 
         return NULL;
     } else { /* cuckoo insert current won't fail as long as size is valid */
-        it = cuckoo_insert(key, NULL, time_reltime(0));
+        /* Currently, expiry is set as a big number as a workaround to mean
+           never expire */
+        it = cuckoo_insert(key, NULL, INT32_MAX - 1);
```
thinkingfish: Not your change- but I wonder if it's better to just set it to INT32_MAX, but maybe it's not that important...

src/util/define.h
kevyang: thanks for catching that, removed it.

src/server/slimredis/data/cmd_bitmap.c
kevyang: filed #189 to look into this.

Issue #189, opened 2018-07-11
Look into using INT32_MAX for expiry


kevyang:
Currently, we use INT32_MAX - 1 rather than just INT32_MAX for setting expiry for items that have infinite TTL. It is not really clear why we don't just use INT32_MAX; this may be an artifact of porting over from memcache.

PR #193, opened 2018-07-25
Use INT32_MAX for unbounded expiry

kevyang:
Addresses #189. I believe using UINT32_MAX - 1 was in reaction to cuckoo, where we assumed that timestamp will be less than UINT32_MAX for the sake of expiry based eviction. This change updates the behavior to use INT32_MAX to mean unbounded expiry in all places, and slightly changes the behavior of cuckoo's selection for eviction when using CUCKOO_POLICY_EXPIRE policy when multiple items share the min expiry timestamp. The new behavior makes it so that instead of selecting the first one in this case, the last one will be selected since we now select if expiry is <= to the min we've inspected, rather than <. I don't think this should cause any issue.

### DIFF
--- a/src/storage/cuckoo/cuckoo.c
+++ b/src/storage/cuckoo/cuckoo.c
@@ -100,7 +100,11 @@ _select_candidate(const uint32_t offset[])
     if (cuckoo_policy == CUCKOO_POLICY_RANDOM) {
         selected = offset[RANDOM(D)];
     } else if (cuckoo_policy == CUCKOO_POLICY_EXPIRE) {
-        proc_time_i expire, min = UINT32_MAX; /* legal ts should < UINT32_MAX */
+        /*
+         * Selection prefers the item expiring soonest, followed by the one
+         * with the smallest offset.
+         */
+        proc_time_i expire, min = INT32_MAX;
         uint32_t i;
 
         for (i = 0; i < D; ++i) {
--- a/src/time/time.h
+++ b/src/time/time.h
@@ -239,7 +239,7 @@ static inline proc_time_i
 time_memcache2proc_sec(memcache_time_u t)
 {
     if (t == 0) {
-        return INT32_MAX - 1;
+        return INT32_MAX;
     }
 
     if (t > TIME_MEMCACHE_MAXDELTA_SEC) {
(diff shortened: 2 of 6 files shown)


src/time/time.h
```diff
@@ -239,7 +239,7 @@ static inline proc_time_i
 time_memcache2proc_sec(memcache_time_u t)
 {
     if (t == 0) {
-        return INT32_MAX - 1;
+        return INT32_MAX;
```
thinkingfish: For consistency should we use `INT64_MAX` instead of `INT64_MAX - 1` in the finer grain timestamps?

src/storage/cuckoo/cuckoo.c
```diff
@@ -100,12 +100,12 @@ _select_candidate(const uint32_t offset[])
...
         selected = offset[RANDOM(D)];
     } else if (cuckoo_policy == CUCKOO_POLICY_EXPIRE) {
-        proc_time_i expire, min = UINT32_MAX; /* legal ts should < UINT32_MAX */
+        proc_time_i expire, min = INT32_MAX; /* legal ts should <= INT32_MAX */
         uint32_t i;
 
         for (i = 0; i < D; ++i) {
             expire = item_expire(OFFSET2ITEM(offset[i]));
-            if (expire < min) {
+            if (expire <= min) {
```
thinkingfish: if `ttl = 0` gets translated to `INT32_MAX`, we can assume that any expiry less than that is an actual expiry instead of an approximation for "never expire". Since we start by "selecting" the first offset, in the case where no item has a shorter TTL, we will still go with the first offset, regardless of the value chosen to indicate max TTL. So my understanding is you can keep the current code unchanged and it will still work with the new init value of `min`, and it will retain the behavior of evicting the first item if there's nothing expiring sooner, is that correct?

kevyang: hm, I think you are correct. we don't need to make an exception for unbounded timestamp, i think i was thrown off by the "legal timestamp" comment. cool, i'll change this back to the original behavior.

src/time/time.h
kevyang: yes, thanks for reminding me. done.

kevyang:
> Also, could you update server/slimredis/data/data/cmd_bitmap.c, line 81 to be consistent with the use of cuckoo timestamp in slimcache?

I think this is done, did you have something else in mind?

PR #194, opened 2018-07-25
Omnibus CDB improvements

slyphon:
This is a rollup of a bunch of work done on top of the ccommon rust changes:

* clean up a bunch of lint errors
* move to using failure crate
* cdb-generation-utility for running performance tests
* selectable mmap or heap backends
* bindgen the cdb_rs.h include file so we can share structures between C and Rust
* merged cdb_ffi and cdb_rs


src/server/cdb/data/process.c
```diff
@@ -68,7 +68,7 @@ process_teardown(void)
     }
 
     if (cdb_handle != NULL) {
-        struct CDBHandle *p = cdb_handle;
+        struct cdb_handle *p = cdb_handle;
         cdb_handle = NULL;
         cdb_handle_destroy(p);
```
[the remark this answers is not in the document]
slyphon: good catch!

src/server/cdb/main.c
```diff
@@ -80,17 +80,24 @@ teardown(void)
...
     cdb_setup();
 
-    char *cdb_file_path = opt->cdb_file_path.val.vstr;
-    if (cdb_file_path == NULL) {
+    struct cdb_handle_create_config cfg;
+    bstring_init(cfg.path);
+
+    cfg.load_method = (opt->use_mmap.val.vbool) ? CDB_MMAP : CDB_HEAP;
+
+    bstring_set_text(cfg.path, opt->cdb_file_path.val.vstr);
```
kevyang: have you tested this? i think `bstring_set_text` only works for string literals

src/server/cdb/data/process.c
```diff
@@ -68,7 +68,7 @@ process_teardown(void)
...
     }
```
kevyang: adding these comments here, because github doesn't support inline comments outside of the diff:

line 78: we can just use bstring_init to reset value_buf
like 104: nit: i think you can just say `rsp->vstr = *vstr;`

src/server/cdb/main.c
slyphon: i think `bstring_set_raw` is for string literals. I've started up the server with this code and it answers queries.

src/server/cdb/data/process.c
slyphon: ah, good call!

I wasn't 100% sure about the semantics for doing that assignment on 104 so i just did the safe thing :)

src/server/cdb/data/process.c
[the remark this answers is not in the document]
slyphon: ok, i fixed that in the latest revision and added a test

src/server/cdb/main.c
kevyang: whoops, you are right. i misread.

slyphon: i had to try both :)

src/storage/cdb/scripts/config.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+# This configuration is used by build and vendorizing scripts
+
+export PROJECT_NAME="cdb_rs"
+export TOOLCHAIN_VERSION="dev"
+export VENDORED_SOURCES_SUFFIX="thirdparty"
+export VENDORED_SOURCES_PACKER_ROLE="io-perf"
```
thinkingfish: leaking Twitter-specific concept such as packer and service role here

src/storage/cdb/scripts/vendorize.sh
```diff
@@ -0,0 +1,47 @@
...
+mkdir -p dist
+
+TARBALL="dist/${PROJECT_NAME}"-"${VENDORED_SOURCES_SUFFIX}.tgz"
+
+# create archive
+tar -C "$TEMP" -czf "$TARBALL" .cargo vendor
+
+# next-steps
+echo 'upload to packer with:'
+echo "package upload -r ${VENDORED_SOURCES_PACKER_ROLE} ${PROJECT_NAME}-${VENDORED_SOURCES_SUFFIX} ${TARBALL}"
```
thinkingfish: ditto

src/storage/cdb/cdb_rs/src/cdb/backend.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+
+impl AsRef<[u8]> for Backend {
+    fn as_ref(&self) -> &[u8] { &self[..] }
+}
+
+impl Backend {
+    pub fn noop() -> Result<Backend> {
+        let v = {
+            let mut buf  = Vec::with_capacity(2048);
```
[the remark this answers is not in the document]
slyphon: non-arbitrary, that's how big the initial (primary) index size is. I'll use the constant I defined elsewhere, I just got lazy.

src/storage/cdb/scripts/config.sh
slyphon: I mean, i'm open to suggestions. Seems like kind of a hassle to have this stored separately.

One thing we could do to keep this out of master would be to put it in an orphan branch (like `gh-pages`). That way it's still "in the repo" but it's not part of the codebase.

ci/local-clean-build-and-test.sh
```diff
@@ -0,0 +1,20 @@
...
+TEMP="$(mktemp -d -t TEMP.XXXXXXX)" || die "failed to make tmpdir"
+cleanup() { [[ -n "${TEMP:-}" ]] && rm -rf "${TEMP}"; }
+trap cleanup EXIT
+
+
+if [[ -n "$(git status --porcelain)" ]]; then
+  die "dirty working copy state, please commit changes before running this script"
+fi
+
+git archive --format=tar --prefix=pelikan/ HEAD .|tar -C "$TEMP" -xf-
```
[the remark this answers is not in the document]
slyphon: [putting prosthetic neckbeard on top of actual neckbeard]

Most shell I've seen does _not_ put spaces around pipes, and I never do. It's not wrong to put spaces around the pipe (and I think some prefer that convention) but I always thought it's kinda amateurish.

src/server/cdb/main.c
slyphon: some suggestions:

`bstring_set_raw` might be better named `bstring_set_literal`
`bstring_set_text` might be better named `bstring_set_char_p` or `bstring_from_char_ptr`

kevyang: I like `bstring_set_literal`. I think `bstring_set_cstr` is a clear name to me, not sure how you guys feel about it though.

slyphon: `bstring_set_cstr` is good. +1

Rust has `CStr`/`CString` to refer to a null-terminated string, so it follows that naming convention well.

thinkingfish: Created https://github.com/twitter/ccommon/issues/175 to update APIs as suggested

ci/install-check.sh
```diff
@@ -17,7 +17,9 @@ TEMP="$(mktemp -d -t TEMP.XXXXXXX)" || die "failed to make tmpdir"
 cleanup() { [[ -n "${TEMP:-}" ]] && rm -rf "${TEMP}"; }
 trap cleanup EXIT
 
-TOPLEVEL="$(git -C "$(cd "$(dirname "$0")" >/dev/null || exit 1; pwd)" rev-parse --show-toplevel)" || die 'failed to find TOPLEVEL'
+realpath() { python -c "import os,sys; print os.path.realpath(sys.argv[1])" "$1"; }
```
thinkingfish: So this assumes python directs to python2. Probably fine for now, just want to point out.

slyphon: ugh, ok you're right. i'll fix it

slyphon: fixed in #199

PR #201, opened 2018-08-03
add some redis list commands

kevyang:
- added support for some redis list commands
- expanded ziplist API to contain trim, and added unit tests
- some minor item API additions for slabbed storage


kevyang:
@thinkingfish wanted some input as to what the behavior should be for list find command, i have a comment in the diff that summarizes my uncertainty. still need to flesh out list get more to support multi get, and also need to write list push. will get working on these soon.

src/storage/slab/item.c
```diff
@@ -128,6 +128,12 @@ _item_link(struct item *it)
     PERSLAB_INCR_N(it->id, item_val_byte, it->vlen);
 }
 
+bool
+item_will_fit(const struct item *it, uint32_t len)
+{
+    return slab_id(item_size(it)) == item_slabid(it->klen, it->vlen + len, it->olen);
```
thinkingfish: I think it's cheap if we:
- inline this function;
- make (slab) profile visible (so we can define an inline function in the header);
- implement as `return (profile[it->id] <= item_size(it) + len);`

Mostly this is cheaper because `item_slabid` does a binary search through `profile` array.

thinkingfish: Also, maybe changing `len` something like `delta` or `increment` will make the purpose of the function even more obvious?

src/data_structure/ziplist/ziplist.h
```diff
@@ -246,6 +246,17 @@ ziplist_rstatus_e ziplist_remove_val(uint32_t *removed, ziplist_p zl, const stru
...
+ * trim list down to contain at most count entries, starting at idx. a negative count means
+ * starting from the end, and gives a list that is non inclusive of idx.
+ *
+ * e.g.
+ * if list contains { 0, 1, 2, 3, 4, 5 }, and we call ziplist_trim(zl, 4, -3),
+ * the trimmed list will be { 1, 2, 3 }.
+ *
+ * if there are fewer than count entries, all entries starting at idx are preserved
+ */
+ziplist_rstatus_e ziplist_trim(ziplist_p zl, int64_t idx, int64_t count);
```
thinkingfish: What do you think the relationship between `ziplist_remove` and `ziplist_trim` should be? I admit I haven't thought about this very carefully, but do you think trim will either enable or simplify implementation of commands that remove doesn't support well?

thinkingfish:
Maybe we can shelf/remove `list_find` given redis doesn't really have this API? If nobody has asked for it I wonder if people using lists simply don't need to find by value very often.

If we have a case for "find", I think we may also need to clarify the API to mean "find the first occurrence of X", and maybe even change the syntax to be `List.find KEY VALUE [INDEX]` so we can find subsequent occurrences? Anyway seems too many uncertainties to worry about for now.

src/data_structure/ziplist/ziplist.h
kevyang: I probably haven't thought about this carefully enough either. The most obvious benefit of `ziplist_trim` as currently implemented right now is it saves on one remove call if we are trimming in the middle of the list. But for other cases, it is essentially the same as remove. Having this API obviously simplifies trim commands, but I can't think of too many other uses for it at the moment.

src/data_structure/ziplist/ziplist.c
```diff
@@ -353,12 +389,11 @@ ziplist_remove_val(uint32_t *removed, ziplist_p zl, const struct blob *val,
         return ZIPLIST_EINVALID;
     }
 
-    if (count == 0) {
-        return ZIPLIST_EINVALID;
-    }
-
+    count = count == 0 ? INT64_MAX : count;
```
thinkingfish: I feel this is probably me not explaining what I was thinking when implementing this function- I was leaning toward leaving the translation ("0 -> all") to the upper layer, similar to how we handled TTL timestamps in the protocol/storage, to keep the lower levels more straightforward. How do you feel about this?

src/data_structure/ziplist/ziplist.c
```diff
@@ -353,12 +389,11 @@ ziplist_remove_val(uint32_t *removed, ziplist_p zl, const struct blob *val,
...
     atmost = forward ? count : -count;
-    *removed = 0;
+    if (removed != NULL) {
```
thinkingfish: Good catch guarding against empty pointer. Shall we use an internal variable and assign that to `*removed` at the end, to avoid extra jump instructions?

src/data_structure/ziplist/ziplist.c
kevyang: yea, i actually like the original behavior better too. the reason i changes this is that count == 0 for lrem in redis means remove all. i will revert this; perhaps we can move this logic out to protocol if we want to support the redis behavior.

kevyang: however, returning error also seems wrong to me. intuitively, it seems like this case should just be a no op

thinkingfish: Agree that noop is the better behavior.

This file is part of "data_structure", we can still put the translation in Redis' lrem implementation, e.g. in `src/server/redis/data/cmd_list.c` or somewhere like that.

PR #202, opened 2018-08-07
rename the redis binaries to ds (dataserver) 

thinkingfish:
I started introducing different APIs based on the RESP so these should not be called Redis to avoid confusion.

This is the first of the two PRs needed for the clean-up, which only affects the code under `src/server`.

The changes under `src/protocol` is more involved, because we are essentially introducing two sets of APIs that share the same wire protocol. I need to think about how to minimize duplicate code while allowing both to be compiled simultaneously (which may or may not be feasible when implemented in C).


kevyang:
i am fine with this for now, but as we finalize the protocol we may want to think of a less generic name than "dataserver"

thinkingfish:
@kevyang agree. Once we have a name for the protocol we will rename again.

PR #206, opened 2018-11-30
adding script to generate (stacked) server instances on a single host

thinkingfish:
Example Usage:
` python server_config.py --instances 30 --prefix test00`


scripts/load_testing/server_config.py
```diff
@@ -0,0 +1,100 @@
...
+daemonize: yes
+admin_port: {admin_port}
+server_port: {server_port}
+
+buf_init_size: 4096
+
+buf_sock_poolsize: 16384
+
+debug_log_level: 5
+debug_log_file: log/{server_port}/twemcache.log
```
HarpPDX: Need to add prefix before log paths in config.

scripts/load_testing/client_config.py
```diff
@@ -0,0 +1,82 @@
...
+  with open(os.path.join(prefix, 'rpcperf.toml'), 'w') as the_file:
+    the_file.write(config_str)
+
+def generate_runscript(prefix, instances):
+# create test.sh
+  fname = os.path.join(prefix, 'test.sh')
+  with open(fname, 'w') as the_file:
+    for i in range(instances):
+      server_port = PELIKAN_SERVER_PORT + i
+      the_file.write('{binary_file} --config {config_file}'.format(binary_file=RPCPERF_BINARY, config_file='rpcperf.toml'))
```
HarpPDX: Prefix on path to toml woud make it so we can run from the base directory.

scripts/load_testing/client_config.py
```diff
@@ -0,0 +1,82 @@
...
+      the_file.write(' --server {server_ip}:{server_port} &\n'.format(server_ip=PELIKAN_SERVER_IP, server_port=server_port))
```
HarpPDX: ?? redirect output to instance-specific output file, e.g. 
     the_file.write('{binary_file} --config {prefix}/{config_file}'.format(binary_file=RPCPERF_BINARY, prefix=prefix, config_file='rpcperf.toml'))
      the_file.write(' --server {server_ip}:{server_port} 2>&1'.format(server_ip=PELIKAN_SERVER_IP, server_port=server_port))
      the_file.write(' > {prefix}/rpc_perf_{server_port}.log'.format(server_port=server_port, prefix=prefix))
      the_file.write(' &\n')

scripts/load_testing/server_config.py
thinkingfish: Shall we assume we launch `bring-up.sh` from the test directory, e.g. `test_32_4`, instead of one level up? This makes the test folder self-contained, i.e. `bring-up.sh` assumes `pwd` is where itself resides, and all paths are relative to that.

This actually means we should _remove_ prefix from the `config_path` parameter below. And do something like the following to launch an experiment
```
cd pelikan_32_4 && bring-up.sh && cd ..
cd rpcperf_100_32_4 && run.sh && cd ..
```

scripts/load_testing/client_config.py
```diff
@@ -0,0 +1,82 @@
...
+request-timeout = 200
+connect-timeout = 50
+
+[[workload]]
+name = "get"
+method = "get"
+rate = {rate_get}
+  [[workload.parameter]]
+  style = "random"
+  size = {size}
```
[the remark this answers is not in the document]
thinkingfish: I'm doing some calculation on how to properly config keyspace for different sizes, will address your client-side comments later.

scripts/load_testing/server_config.py
HarpPDX: Sounds good.

thinkingfish:
@HarpPDX I have updated the scripts to use server-side warmup/prefill. The options for `client_config.py` have slight updates too, what's worth noting is that I'm treating `connections` as the total number of connections for each rpcperf process, not per thread, and handle the connection splitting internally in the script.

`generate.sh` is now updated to produce all the configs, and I have removed the `key+val=32B` workload from the test plan because I think it's a good thing to use 32-bit key length throughout the tests. So we will have 24 server config plans and 72 client config plans in total.

scripts/load_testing/client_config.py
```diff
@@ -0,0 +1,82 @@
...
+  config_str = """\
+[general]
+threads = {threads}
+tcp-nodelay = true
+connections = {connections}
+windows = 600
+duration = 60
+request-timeout = 200
+connect-timeout = 50
+
```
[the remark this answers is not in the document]
thinkingfish: warmup is now moved into server config

scripts/load_testing/client_config.py
thinkingfish: I will address this shortly

scripts/load_testing/client_config.py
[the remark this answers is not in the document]
thinkingfish: key size is now fixed at 32 and value size makes up for the rest of the payload size.

thinkingfish:
Here's the default output of `client_config.py`
```
$ cat loadgen/rpcperf.toml
[general]
threads = 1
tcp-nodelay = true
connections = 100
windows = 600
duration = 60
request-timeout = 200
connect-timeout = 50

[[workload]]
name = "get"
method = "get"
rate = 9000
  [[workload.parameter]]
  style = "random"
  size = 32
  num = 38347923
  regenerate = true

[[workload]]
name = "set"
method = "set"
rate = 999
  [[workload.parameter]]
  style = "random"
  size = 32
  num = 38347923
  regenerate = true
  [[workload.parameter]]
  style = "random"
  size = 32
  regenerate = false

$ cat loadgen/test.sh
/root/Twitter/rpc-perf/target/release/rpc-perf --config rpcperf.toml --server 10.25.2.44:12300 &
/root/Twitter/rpc-perf/target/release/rpc-perf --config rpcperf.toml --server 10.25.2.44:12301 &
/root/Twitter/rpc-perf/target/release/rpc-perf --config rpcperf.toml --server 10.25.2.44:12302 &
```

Here are the default output or `server_config.py`:
```
]$ cat test/bring-up.sh
sudo numactl --cpunodebind=0 --preferred=0 /root/Twitter/pelikan/_build/_bin/pelikan_twemcache config/pelikan-12300.config
sudo numactl --cpunodebind=1 --preferred=1 /root/Twitter/pelikan/_build/_bin/pelikan_twemcache config/pelikan-12301.config
sudo numactl --cpunodebind=0 --preferred=0 /root/Twitter/pelikan/_build/_bin/pelikan_twemcache config/pelikan-12302.config

$ cat test/config/pelikan-12300.config
daemonize: yes
admin_port: 9900
server_port: 12300

buf_init_size: 4096

buf_sock_poolsize: 16384

debug_log_level: 5
debug_log_file: log/12300/twemcache.log
debug_log_nbuf: 1048576

klog_file: log/12300/twemcache.cmd
klog_backup: log/12300/twemcache.cmd.old
klog_sample: 100
klog_max: 1073741824

prefill: yes
prefill_ksize: 32
prefill_vsize: 32
prefill_nkey: 38347923

request_poolsize: 16384
response_poolsize: 32768

slab_evict_opt: 1
slab_prealloc: yes
slab_hash_power: 26
slab_mem: 4294967296
slab_size: 1048756

time_type: 2
```

scripts/load_testing/client_config.py
```diff
@@ -0,0 +1,97 @@
...
+  config_str = """\
+[general]
+threads = {threads}
+tcp-nodelay = true
+connections = {connections}
+windows = 600
+duration = 60
+request-timeout = 200
+connect-timeout = 50
```
brayniac: 50ms connect timeout? Seems aggressive compared to the request timeout.

scripts/load_testing/generate.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+conn_configs=(100 1000 10000)
+mem_configs=(4 8 16 32)
+size_configs=(64 128 256 512 1024 2048)
+instances=30
+ksize=32
+rate=100000
+threads=2
```
brayniac: Remember that threads is a multiplier on connections. The `--connections` param for rpc-perf is actually the per-endpoint poolsize for each client thread. I think in the rpc-perf configs section you'd want to take the ceiling of the connections divided by the rpc-perf threads

scripts/load_testing/client_config.py
thinkingfish: 200/200?

scripts/load_testing/client_config.py
```diff
@@ -0,0 +1,97 @@
...
+  rate_set = int(rate * RPCPERF_SET_RATIO)
+  nkey = int(ceil(1.0 * slab_mem / (vsize + KSIZE + PELIKAN_ITEM_OVERHEAD)))
+  conn_per_thread = connections / threads
+
+  config_str = """\
+[general]
+threads = {threads}
+tcp-nodelay = true
+connections = {connections}
+windows = 600
```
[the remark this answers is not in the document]
thinkingfish: Good catch

scripts/load_testing/generate.sh
thinkingfish: Yep, that's what I'm doing right now: https://github.com/twitter/pelikan/blob/9f791895c494a51eb5c252587c21dc00af72923d/scripts/load_testing/client_config.py#L27 and https://github.com/twitter/pelikan/blob/9f791895c494a51eb5c252587c21dc00af72923d/scripts/load_testing/client_config.py#L62

thinkingfish:
@HarpPDX changes since yesterday:

instead of having each pelikan using its own log directory under `log`, e.g. `log/12300/`, put all log files under one directory, using port number in the filename to differentiate. This makes grepping a little easier. There are some other minor naming changes too.

added `runtest.sh` to allow launching test from the client side, which first launch the server instances remotely (via ssh) on the target machine, waiting for warmup to complete; then launch clients; then remotely kill the server instances. Looks like between the sandboxes `ssh` doesn't require human input, so hopefully this makes it a little less tedious when we are ready to run a series of tests in a row.

HarpPDX:
@thinkingfish Looks good. Keep in mind we may need to run the client-side on multiple machines to fully load the server. We can cross that bridge when we get to it.

scripts/load_testing/client_config.py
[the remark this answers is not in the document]
HarpPDX: FYI...
windows = 1 and duration = 600 ... results in a 20 minute total runtime.

scripts/load_testing/client_config.py
```diff
@@ -0,0 +1,88 @@
...
+windows = 2
+interval = 60
+request_ratelimit = {rate}
+
+[[keyspace]]
+length = 32
+count = {nkey}
+weight = 1
+commands = [
+    {{action = "get", weight = 9}},
```
brayniac: Weights should be calculated from the get/set ratio (or those parameters converted to weight instead of ratio)

scripts/load_testing/client_config.py
```diff
@@ -0,0 +1,88 @@
...
+clients = {threads}
+tcp-nodelay = true
+poolsize = {connections} # this specifies number of connection per thread
+# runtime ~= windows x duration
+windows = 2
+interval = 60
+request_ratelimit = {rate}
+
+[[keyspace]]
+length = 32
```
brayniac: this should be taken from the KSIZE

scripts/load_testing/client_config.py
thinkingfish: switched to explicit weight.

PR #208, opened 2018-12-18
add prefill option and logic to populate the slabs with data

thinkingfish:
This allows us to quickly fill a slab's heap with printed numeric keys/values, this is useful for performance tests.

Cache's RSS, which is primarily driven by the amount data stored in slab, is an important factor for performance. As memory gets bigger on modern hardware, it takes longer to fill the heap by external traffic. For example, if we want to warm up a 32GB heap, using 64 bit key+val sizes, we will need to create 300M keys. If we use 100K qps to warm it up, it will take almost an hour. This clearly makes realistic load testing fairly slow and expensive to conduct. In comparison, prefilling the memory inside the server can make this process at least an order of magnitude faster.

The implementation introduces a master switch `prefill`, and 3 options that control how prefilling should be performed, regarding key size, value size, and number of keys inserted. It also provides a summary of the operation including timing in the log.

### DIFF
--- a/src/server/twemcache/data/process.c
+++ b/src/server/twemcache/data/process.c
@@ -25,6 +25,41 @@ typedef enum put_rstatus {
 static bool process_init = false;
 static process_metrics_st *process_metrics = NULL;
 static bool allow_flush = ALLOW_FLUSH;
+static bool prefill = PREFILL;
+static uint32_t prefill_ksize;
+static char prefill_kbuf[UINT8_MAX]; /* slab implementation has klen as unint8_t */
+static uint32_t prefill_vsize;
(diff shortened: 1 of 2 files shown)


thinkingfish:
Example:

With config:
```
prefill: yes
prefill_ksize: 5
prefill_vsize: 10
prefill_nkey: 16
```

We can verify the keys:
```
get 00000
VALUE 00000 0 10
0000000000
END
get 00009
VALUE 00009 0 10
0000000009
END
get 00015
VALUE 00015 0 10
0000000015
END
get 00016
END
```

Summary in log looks like this:
```
prefilling slab with 10000000 keys, of key len 32 & val len 32, in 19.424 seconds
```

PR #216, opened 2019-02-08
cmake: fix check detection

pbalcer:
Building on Ubuntu 18.04 fails with missing symbols message
due to lack of proper libraries being automatically linked
against tests.

This patch uses pkgconfig to automatically detect system-installed
libcheck, and also gets rid of the unnecessery symbol requirement in
favor of a simple version check.


pbalcer:
I'm not sure if this is correct in the context of your build system, but that's a common approach for searching for dependencies. I'm also not sure why you have two separate cmake projects doing the same thing - but I retained did retain that.

PR #217, opened 2019-02-15
Initial implementation of file-backed durable cuckoo storage engine

pbalcer:


pbalcer:
This is the easy part of pmem enabling ;) I made some simple tests and ran a quick benchmark, and things look OK.

Next, I'll create a PR with the micro benchmark for storage engines, and then move on to slab/hashtable/item modifications.

src/datapool/datapool_pmem.c
```diff
@@ -0,0 +1,217 @@
...
+ * The the datapool to retain its contents, the datapool_close() call must
+ * finish successfully.
+ */
+struct datapool *
+datapool_open(const char *path, size_t size, int *fresh)
+{
+    struct datapool *pool = cc_alloc(sizeof(*pool));
+    if (pool == NULL) {
+        log_error("unable to create allocate memory for pmem mapping");
+        goto err_alloc;
```
kevyang: wonder if using `goto` really makes sense here; each `goto` label only has one `goto` sending it there

src/datapool/datapool_pmem.c
```diff
@@ -0,0 +1,217 @@
...
+        goto err_map;
+    }
+
+    log_info("mapped datapool %s with size %llu, is_pmem: %d",
+        path, pool->mapped_len, pool->is_pmem);
+
+    pool->hdr = pool->addr;
+    pool->user_addr = (uint8_t *)pool->addr + sizeof(struct datapool_header);
+
+    if (fresh)
```
kevyang: nit: please put brackets around conditionals, even if it's just a single statement. there are a few instances of this

src/datapool/datapool_pmem.c
```diff
@@ -0,0 +1,217 @@
...
+}
+
+void
+datapool_close(struct datapool *pool)
+{
+    datapool_sync(pool);
+    datapool_flag_clear(pool, DATAPOOL_FLAG_DIRTY);
+
+    if (pool->file_backed)
+        pmem_unmap(pool->addr, pool->mapped_len);
```
kevyang: interesting that there are no failure scenarios for these pmem lib functions. will have to read more about how they work :)

test/CMakeLists.txt
```diff
@@ -7,5 +7,8 @@ add_subdirectory(protocol)
 add_subdirectory(hotkey)
 add_subdirectory(storage)
 add_subdirectory(time)
+if(USE_PMEM)
```
kevyang: kind of confused as to what happens when we run cmake without USE_PMEM. would have thought this wouldn't work due to cuckoo needing datapool stuff but maybe i am misunderstanding.

src/datapool/datapool_pmem.c
```diff
@@ -0,0 +1,217 @@
...
+#include <cc_mm.h>
+#include <cc_debug.h>
+
+#include <inttypes.h>
+
+#include <libpmem.h>
+
+#define DATAPOOL_SIGNATURE ("PELIKAN") /* 8 bytes */
+#define DATAPOOL_SIGNATURE_LEN (sizeof(DATAPOOL_SIGNATURE))
+#define DATAPOOL_HEADER_LEN 4096
```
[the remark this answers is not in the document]
pbalcer: Do you expect to run Pelikan on architectures that do not support 4k pages? :)

But no, there no correctness (or otherwise) reasons for this number. It's just big enough for any future expansion.

Also, msync() doesn't have to internally align the size if we pass it a nice and round number.

src/datapool/datapool_pmem.c
```diff
@@ -0,0 +1,217 @@
...
+    pmem_msync(pool->hdr, DATAPOOL_HEADER_LEN);
+}
+
+static void
+datapool_sync(struct datapool *pool)
+{
+    pmem_msync(pool->addr, pool->mapped_len);
+}
+
+static int
```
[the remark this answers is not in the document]
pbalcer: Done.

src/datapool/datapool_pmem.c
pbalcer: Done

src/datapool/datapool_pmem.c
pbalcer: `msync` and `munmap` can only realistically fail on invalid arguments (misaligned address being the most common error). The `pmem_map_file` error path is covered.

test/CMakeLists.txt
pbalcer: By default the pmem implementation of the data pool is not compiled, and instead the shm (`malloc`/`free`) implementation is used as a fallback. What this means is that there's no change whatsoever to the way Pelikan functions, everything will work just the way it used to. Only if it's built with PMEM support explicitly enabled and libpmem installed, the pmem implementation of the datapool abstraction is going to be compiled in, and the `cuckoo_datapool` will be taken into account when starting the cache.

This is implemented in `src/datapool/CMakeLists.txt`.

src/datapool/datapool_pmem.c
pbalcer: This is a common error handling idiom when cleanup is needed before exiting the function. There should be only one `goto` per label because this allows for stacking the cleanup functions. You could say that this is poor's man RAII.

src/datapool/datapool_pmem.c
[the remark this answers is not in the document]
kevyang: got it. we don't plan to use such architectures afaik (though we have evaluated the effect of hugepage before), but this is an open source project so ideally we would like to support a multitude of use cases if we can. 4 KB seems quite large to me but considering the amount of memory we are dealing with using pmem, this seems fine to me.

src/datapool/datapool_pmem.c
kevyang: cool, i'm fine with leaving this as is. it isn't explicitly out of line with our style guide, and i don't think it really has a negative impact on readability; i just haven't seen this idiom before.

PR #219, opened 2019-05-22
Fix cleaning after partial request

jschmieg:
Using partial request, like from rpc-perf, causes first request
buffer not to be returned to requests pool.

This commits adds function which will return this buffer to pool.

### DIFF
--- a/src/server/slimcache/data/process.c
+++ b/src/server/slimcache/data/process.c
@@ -448,6 +448,12 @@ _cleanup(struct request **req, struct response **rsp)
     response_return_all(rsp);
 }
 
+static inline void
+_cleanup_req(struct request **req)
+{
+    request_return(req);
+}
+
 int
 slimcache_process_read(struct buf **rbuf, struct buf **wbuf, void **data)
 {
@@ -480,6 +486,7 @@ slimcache_process_read(struct buf **rbuf, struct buf **wbuf, void **data)
         if (status == PARSE_EUNFIN || req->partial) { /* ignore partial */
             (*rbuf)->rpos = old_rpos;
             buf_lshift(*rbuf);
+            _cleanup_req(&req);
             return 0;
         }
         if (status != PARSE_OK) {


thinkingfish:
Does rpc-perf not follow up with the rest of the request later? Trying to fully understand the behavior here.

We keep the partial request around to avoid repeat the parsing already done.

jschmieg:
In case of partial request, processing of the first part of request may allocate memory in slimcache_process_read(): 
req = request_borrow(); 
and not return it properly, which is causing a leak. We ended up with OOM during longer rpc-perf benchmarks. This PR corrects that.

thinkingfish:
I understand the intention, I was trying to figure out if this is the best/right place for a fix. Actually, after going back to the code, I think the best fix should be somewhere else.

First- the idea here is that as long as a connection is considered "live", we let it hold onto at most one request object to avoid having to borrow/return every time the read event happens. The request object should be returned upon connection termination.

In the `core/data/worker.c` code, an application specific function is called (https://github.com/twitter/pelikan/blob/master/src/core/data/worker.c#L156) upon error events on the socket (https://github.com/twitter/pelikan/blob/master/src/core/data/worker.c#L214). In slimcache, this function is implemented as `slimcache_process_error` in https://github.com/twitter/pelikan/blob/master/src/server/slimcache/data/process.c#L571 . Compared to twemcache, it's missing the logic to return the request objects, which in twemcache is implemented at https://github.com/twitter/pelikan/blob/master/src/server/twemcache/data/process.c#L841 

This is my theory. Could you confirm whether twemcache exhibit this leaky behavior or not? If twemcache behaves normally, I think we should add the corresponding logic to `slimcache_process_error` as a fix.

michalbiesek:
@thinkingfish 
I was investigate this issue. Please see my conclusions below:

> This is my theory. Could you confirm whether twemcache exhibit this leaky behavior or not? If twemcache behaves normally, I think we should add the corresponding logic to slimcache_process_error as a fix.

Twemcache doesn't exhibit this leaky behavior.

Regarding the solution you propose, please be aware that slimcache currently got different logic to handle requests than twemcache:

https://github.com/twitter/pelikan/blob/7e4ef09899724d922273f29a689cb237960c066a/src/server/slimcache/data/process.c#L450-L454

Slimcache doesn't hold one request object for whole "connection" - in other words every command sended to slimcache generate new  request borrow/return sequence.

Regarding moving twemcache solution to slimcache I think it is worth to mention that slimcache doesn't have method which allows reserve an item e.g. item_reserve - so IMHO this change won't be so trivial.

thinkingfish:
@jschmieg @michalbiesek You two are right. Every other path in this function leads to returning the req/rsp objects, this is the simplest fix and consistent with the rest of the code.

I guess it's another question whether we want to make it behave more like Twemcache, but until that turns out to be a differentiator for performance, it is likely not super important. I'll merge this patch.

PR #221, opened 2019-05-30
Initial implementation of file-backed durable slab storage

michalbiesek:
Added support to recreate hashtable after graceful shutdown.


src/datapool/datapool_pmem.c
```diff
@@ -31,11 +34,16 @@
  * opened.
  */
 struct datapool_header {
-    uint8_t signature[DATAPOOL_SIGNATURE_LEN];
-    uint64_t version;
-    uint64_t size;
-    uint64_t flags;
-    uint8_t unused[DATAPOOL_HEADER_LEN - 32];
+    struct {
```
kevyang: it's not really clear to me why these need to be in their own structs, rather than adding a user_data member

src/datapool/datapool_shm.c
```diff
@@ -41,3 +41,14 @@ datapool_size(struct datapool *pool)
     return cc_alloc_usable_size(pool);
 }
 
+void
```
kevyang: would appreciate comments to document why empty/unused functions are there

src/datapool/datapool_pmem.c
pbalcer: And it should be the other way around. The signature and version should be first thing in the file. If only so that magic file isn't excessively complicated.

Now that I think about it, we should also add `user_signature` or a `layout` field, so that when there are two pelikan database files we can programmatically distinguish between them, and refuse to open, e.g., cuckoo hashtable in the slab engine, avoiding a segfault.

deps/ccommon/include/cc_queue.h
```diff
@@ -127,6 +127,7 @@ extern "C" {
  * _INSERT_TAIL         -       -       +       +       +
  * _REMOVE_HEAD         +       -       +       -       -
  * _REMOVE              +       +       +       +       +
+ * _REINIT              -       -       -       +       -
```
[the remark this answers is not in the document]
michalbiesek: Done: twitter/ccommon#193

src/datapool/datapool_pmem.c
michalbiesek: I am working on it.

src/datapool/datapool_shm.c
michalbiesek: Done

src/datapool/datapool_pmem.c
michalbiesek: Done. User signature field was added.

michalbiesek:
I updated this PR:
- Added Unit tests for PMEM
- Added references to ccommon repository (changes which should be included regarding this PR)

I am aware that number of changes is pretty big and could be problematic during code review, so I reorganize the whole Pull Request to sort it by commits.
The review should go more smoothly if you review changes commit by commit.

michalbiesek:
I rebase the branch regarding commit 35b5218

PR #262, opened 2019-09-30
Create pelikan rust bindings

swlynch99:
This is the second of 3 PRs for #260.

It has
- Complete (except for static inline functions) low-level bindings to all pelikan modules.
- Just enough high-level bindings to rewrite the pingserver in rust
- Formats cdb-rs using rustfmt

### DIFF
--- a/src/rust-util/pelikan/src/core/mod.rs
+++ b/src/rust-util/pelikan/src/core/mod.rs
@@ -1,0 +1,159 @@
+// Copyright (C) 2019 Twitter, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Wrapper for the pelikan `core` module.
+
+pub mod admin;
+
+use ccommon::buf::OwnedBuf;
+use ccommon_sys::buf;
+use pelikan_sys::core::data_processor;
+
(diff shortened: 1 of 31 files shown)


src/rust-util/pelikan/src/core/mod.rs
```diff
@@ -0,0 +1,129 @@
...
+use std::ffi::c_void;
+use std::mem::MaybeUninit;
+
+#[repr(C)]
+pub enum DataProcessorError {
+    Error = -1,
+}
+
+/// Methods for handling messages on the data port
+pub trait DataProcessor {
```
swlynch99: I'm not happy with how this turned out and I plan to rewrite this in a follow-on PR

src/storage/cdb/cdb_rs/src/cdb/ffi/mod.rs
```diff
@@ -1,15 +1,14 @@
 use cc_binding as bind;
 use ccommon_rs::bstring::BStr;
-use cdb::{cdb_handle, Reader, Result};
 use cdb;
+use cdb::{cdb_handle, Reader, Result};
```
brayniac: assuming this is rustfmt change, but I'd probably do `cdb::{self, ....}`.

src/rust-util/pelikan/src/core/mod.rs
brayniac: Yeah, this is a little strange. What would we be blocking by rewriting now? Is there anything that blocks doing this better in this PR?

src/rust-util/pelikan/src/core/mod.rs
```diff
@@ -0,0 +1,118 @@
...
+    fn read(
+        rbuf: &mut OwnedBuf,
+        wbuf: &mut OwnedBuf,
+        state: &mut *mut (),
+    ) -> Result<(), DataProcessorError>;
+    fn write(
+        rbuf: &mut OwnedBuf,
+        wbuf: &mut OwnedBuf,
+        state: &mut *mut (),
+    ) -> Result<(), DataProcessorError>;
```
brayniac: newline

PR #265, opened 2019-10-26
Add Rust ping server

swlynch99:
This finally adds the pingserver written in rust that was first written for #260. In the process of doing this I discovered that the cmake rust integration was broken when building static libraries or executables. The cmake changes required to fix that are also bundled in here.

Summary of what was changed:
- Added a rust port of the ping server
- Fixed the rust cmake integration so that it works properly when building executables or static libraries
- Remove the previous rust language support files since using them breaks stuff in hard to debug ways

This PR makes changes to the bundled copy of ccommon, those will be backported to ccommon in a separate PR.


CMakeLists.txt
```diff
@@ -233,8 +231,8 @@ endif()
 # vim:ts=4:sw=4:et
 
 if(HAVE_RUST)
-    enable_language(Rust)
-    include(cmake/CMakeCargo.cmake)
+    # enable_language(Rust)
```
brayniac: Should we remove this line instead of making it a comment?

deps/ccommon/CMakeLists.txt
```diff
@@ -196,7 +196,7 @@ if(CHECK_FOUND)
 endif(CHECK_FOUND)
 
 if(HAVE_RUST)
-    enable_language(Rust)
+    # enable_language(Rust)
```
brayniac: Same - prefer to remove commented code

deps/ccommon/cmake/CMakeCargo.cmake
```diff
@@ -300,7 +325,7 @@ function(cargo_build)
         add_custom_command(
             # Dummy file to ensure that the command always runs
             OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/DOES_NOT_EXIST
-            COMMAND ${CARGO_ENV_COMMAND} ${CARGO_EXECUTABLE} build ${CRATE_ARGS}
+            COMMAND ${CARGO_ENV_COMMAND} cargo build ${CRATE_ARGS}
```
brayniac: What's the reason for this change?

swlynch99: `CARGO_EXECUTABLE` was originally exported by `FindRust.cmake`. Since that's been removed in this PR the variable is no longer set so instead it just uses cargo directly.

PR #274, opened 2019-11-27
Update tokio and use LocalSet for non-Send tasks

swlynch99:
Tokio v0.2.1 adds back the ability to run non-send tasks. This PR removes the hack we had in place to do this and replaces it with LocalSet.

Unfortunately, the LocalSet API currently either panics or hangs when it is dropped with futures that are still live (as would happen when we intercept `SIGTERM` and try to exit). In cases where these currently panic I've left them alone but I've added one `mem::forget` call to avoid an infinite loop.


swlynch99:
Going to hold off on merging this until https://github.com/tokio-rs/tokio/pull/1843 is merged and released. After that goes through we can hopefully remove the hacks from here.

swlynch99:
Currently blocked on tokio-rs/tokio#1885

hawkw:
tokio-rs/tokio#1892 should fix tokio-rs/tokio#1885 (thanks for the report @swlynch99!)

PR #275, opened 2019-11-28
Sorted map

thinkingfish:
Generalizing sorted array a bit to allow fixed size values associated with (numeric) keys. Other than changing the signature of three APIs, and small changes to internal implementations, it looks and works very similar to sorted array.

Once this is merged I'm gonna start adding higher level APIs for SortedMap. Since we can use zero value size as a special case to get sorted array behavior with sorted map, I will probably port the sorted array APIs to use sorted map, and drop the sorted array data structure entirely.

### DIFF
--- a/src/rust-util/pelikan-sys/build.rs
+++ b/src/rust-util/pelikan-sys/build.rs
@@ -636,6 +636,26 @@ fn main() {
     if cfg!(feature = "ds_sarray") {
         gen_ds_sarray();
     }
+    if cfg!(feature = "ds_smap") {
+        print_directives("ds_smap", "data_structure/smap");
+
+        let bindings = builder()
+            .header("../../data_structure/smap/smap.h")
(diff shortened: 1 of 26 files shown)


src/rust-util/pelikan-sys/build.rs
```diff
@@ -295,6 +295,26 @@ fn main() {
...
+
+        let bindings = builder()
+            .header("../../data_structure/smap/smap.h")
+            .whitelist_type("smap_p")
+            .whitelist_type("smap_rstatus_e")
+            .whitelist_function("smap_.*")
+            .whitelist_recursively(false)
+            .derive_default(true)
+            .derive_copy(true)
+            .derive_debug(true)
```
swlynch99: This is already handled by the `builder` function. It's better if we don't override this - especially if we want to change the defaults at a later time. (e.g. one of the PRs I'm working on does exactly this)

PR #276, opened 2019-12-05
Implement a cache server which supports both memcache and HTTP protocols

swlynch99:
This PR is essentially two PRs combined into one (and before proper review happens it should be split out into two)

1. It implements a bunch of required bindings for the `memcached` protocol as well as the `hotkey` and `slab` libraries.
2. It implements a cache server that exposes both an HTTP interface and the twemcache interface.

If you're just interested in knowing how the dual-protocol server works, look at the changes under `src/server/dual-protocol` (which really could use a better name).

Stuff that is currently still TODO:
- [x] Proper licensing on all files 
- [x] Actual tests for the new functionality - the twemcache protocol implementation uses the existing C code from pelikan_twemcache but the HTTP handling code is currently untested. (there is a basic test now - it's probably not enough to be production ready but I don't have time to add more in the next week)
- [x] Separate out bindings change into separate PR so it isn't necessary to review a 3500 line PR. (See #277 for this)
- [ ] Add proper metrics to the HTTP server

Note when reviewing: It should be safe to just ignore the C source under `src/server/twemcache-http/memcache` since it is just a copy of the existing source from twemcache.


src/server/twemcache-http/Cargo.toml
```diff
@@ -0,0 +1,38 @@
...
+clap = "2.33"
+log = "0.4.8"
+thiserror = "1.0.8"
+tokio = { version = "0.2.2", features = [ "net" ] }
+
+rustcore = { path = "../../rustcore" }
+
+httparse = "1.3.4"
+httpdate = "0.3.2"
+httpencode = { git="https://github.com/swlynch99/httpencode" }
```
brayniac: This should probably be moved into this repo.

src/server/twemcache-http/build.rs
```diff
@@ -0,0 +1,74 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::env;
+use std::path::Path;
+
+fn main() {
+    let flag = if cfg!(target_os = "linux") {
+        "OS_LINUX"
+    } else {
```
brayniac: might be better to check if target os matches for mac os and otherwise exit with error

PR #282, opened 2020-01-18
Support resp3 style attributes

thinkingfish:
Main ideas are described in ` src/protocol/data/resp/attribute.md`

Summary  of main changes:
- update token handling to recognize attribute and new null representation, making composite type such as `array` and `attribute` more like normal types (taking up a token slot, parse/compose using the same function, etc);
- adding global attribute list that will be handled by the parser;
- added tests for attribute, and made modifications to recognize above changes;
- adopt OPT_VARIED for command description (borrowed from @swlynch99)

The actual usage of `ttl` and `flag` with the rest of the system (mostly interacting with storage/metadata) will come in a separate diff. This change will not break any existing clients since the higher level logic omit attributes altogether.


src/protocol/data/resp/parse.c
```diff
@@ -81,67 +83,189 @@ _parse_cmd(struct request *req)
...
+
+    /* treat this as it cannot fail, skip anything that is not recognized or has
+     * invalid value so the program will proceed to the request main itself.
+     */
+    if (key->type != ELEM_STR) { /* key has to be a simple string */
+        log_warn("attribute key must be simple string, not type %d", key->type);
+
+        return;
+    }
+    while (++type < ATTRIB_SENTINEL &&
```
kevyang: nit: imo this (and the following if clause) would be more readable as a for

src/protocol/data/resp/parse.c
```diff
@@ -81,67 +83,189 @@ _parse_cmd(struct request *req)
...
+            bstring_compare(&attrib_table[type], &key->bstr) != 0) {}
+    if (type == REQ_SENTINEL) {
```
kevyang: do you mean to be checking against ATTRIB_SENTINEL? if so, can this be caught in the following switch?

src/protocol/data/resp/parse.c
```diff
@@ -81,67 +83,189 @@ _parse_cmd(struct request *req)
...
+
+    /* treat this as it cannot fail, skip anything that is not recognized or has
+     * invalid value so the program will proceed to the request main itself.
+     */
+    if (key->type != ELEM_STR) { /* key has to be a simple string */
+        log_warn("attribute key must be simple string, not type %d", key->type);
+
+        return;
+    }
+    while (++type < ATTRIB_SENTINEL &&
```
kevyang: seems like this is copy pasted from req, maybe we can extract this common logic out into the attribute module?

src/protocol/data/resp/request.h
```diff
@@ -75,17 +75,22 @@ struct command {
...
+    bool                    noreply;/* skip response */
+    bool                    serror; /* server error */
+    bool                    cerror; /* client error */
 
     cmd_type_e              type;
-    struct array            *token;    /* member type: `struct element' */
+    struct array            *token; /* member type: `struct element' */
+
+    /* global attributes */
+    int64_t                ttl;     /* -1 means it is not provided */
```
kevyang: given the INT64_MIN figure given in the reset function, is that what you mean here? the comment says -1

src/protocol/data/resp/parse.c
thinkingfish: yes. fixed.

src/protocol/data/resp/parse.c
thinkingfish: Good call, new local function added.

src/protocol/data/resp/request.h
thinkingfish: Yes- good catch. I originally wanted to use `-1` for not provided, then decided I may want to reserve `-1` (or other small negatives) for some particular meaning in the future, such as infinite or invalid TTL.

src/protocol/data/resp/parse.c
thinkingfish: changed

PR #285, opened 2020-05-05
Sync ccommon and fix the build

brayniac:
Sync ccommon from upstream.

Fixes various places where ccommon rust libraries were renamed or
have interface changes.

Fixes the cmake rust build, where changes in ccommon cmake were
preventing successful build. Copy cargo specific cmake stuff into
pelikan and include before processing sub directories.

Adds the ability to build the rust components with cargo.


brayniac:
CI is failing due to check_pipe in ccommon - likely a timing precision issue with the travis environment.

PR #288, opened 2020-05-18
pingserver-rs rewrite

brayniac:
The old pingserver-rs implementation was centered around tokio and
async style programming. When benchmarking, I found the tokio based
implementation had both elevated and inconsitent latency compared
to the C implementation.

I reimplemented the Rust pingserver to use a mio event loop. When
benchmarking, I found that the performance is similar to the C
implementation in terms of latency characteristics and key systems
performance metrics.


src/server/pingserver-rs/src/worker.rs
```diff
@@ -0,0 +1,181 @@
...
+/// A `Worker` handles events on `Session`s
+pub struct Worker {
+    config: Arc<PingserverConfig>,
+    sessions: Slab<Session>,
+    poll: Poll,
+    receiver: Receiver<Session>,
+}
+
+impl Worker {
+    /// Create a new `Worker` which will get new `Session`s from the MPMC queue
```
thinkingfish, line 14: MPSC queue for now?

I am curious, once this is in good shape, to test out the scalability of multithreaded listener, multithreaded worker, and both.

brayniac: That was a typo, using the MPSC queue from the std library

thinkingfish (approved):

PR #294, opened 2020-08-03
use eventfd for notifications b/w server and worker

noxiouz:
**event_fd** is much faster and easier to use to cross thread/process [notification](https://man7.org/linux/man-pages/man2/eventfd.2.html).
**event_fd** is a basically an atomic uint64 counter, that acts like _pollable_ file descriptor and was designed specifically for notification purposes.

### DIFF
--- a/src/core/data/server.c
+++ b/src/core/data/server.c
@@ -248,11 +248,11 @@ _server_event(void *arg, uint32_t events)
         if (events & EVENT_WRITE) { /* retrying worker notification */
             log_verb("processing server write event on pipe");
             INCR(server_metrics, server_event_write);
-            #ifdef USE_EVENT_FD
+#ifdef USE_EVENT_FD
             _server_notify_worker();
-            #else
+#else
             _server_pipe_write();
-            #endif
+#endif
         }
         if (events & EVENT_ERR) {
             log_debug("processing server error event on pipe");
--- a/src/core/data/server.c
+++ b/src/core/data/server.c
@@ -17,9 +17,17 @@
 #include <string.h>
 #include <sysexits.h>
 
+#ifdef USE_EVENT_FD
+#include <sys/eventfd.h>
+#endif
+
 #define SERVER_MODULE_NAME "core::server"
 
+#ifdef USE_EVENT_FD
+int event_fd_s2w = -1;               /* server(w) -> worker(r) */
+#else
 struct pipe_conn *pipe_new = NULL;   /* server(w) -> worker(r) */
+#endif
 struct pipe_conn *pipe_term = NULL;  /* worker(w) -> server(r) */
 struct ring_array *conn_new = NULL;  /* server(w) -> worker(r) */
 struct ring_array *conn_term = NULL; /* worker(w) -> server(r) */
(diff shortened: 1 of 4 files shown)


noxiouz:
If it works, I guess ring_buf + notify could be wrapped and abstract away as notification queue mechanism

src/core/data/server.c
```diff
@@ -53,6 +61,24 @@ _server_close(struct buf_sock *s)
     buf_sock_return(&s);
 }
 
+#ifdef USE_EVENT_FD
+static inline void
+_server_notify_worker(void)
```
thinkingfish, line 66: How about absorbing _server_pipe_write() into this new function, so the details of eventfd va pipe are concentrated within two functions at most?

src/core/data/server.c
```diff
@@ -63,17 +64,17 @@ _server_close(struct buf_sock *s)
 
 #ifdef USE_EVENT_FD
```
thinkingfish, line 65: this becomes unnecessary given the `#ifdef` in `_server_notify_work()`

src/core/data/server.c
```diff
@@ -96,22 +97,47 @@ _server_pipe_write(void)
...
+    _server_event_fd_write();
+#else
+    _server_pipe_write();
+#endif
+}
+
 /* pipe_read recycles returned streams from worker thread */
 static inline void
-_server_pipe_read(void)
+_server_read_notification(void)
```
thinkingfish, line 112: For naming consistency, shall we call the two notification functions `_server_send_notification` and `_server_read_notification`?

src/core/data/server.c
```diff
@@ -96,22 +97,47 @@ _server_pipe_write(void)
 }
 #endif
 
+static inline void
+_server_notify_worker(void)
+{
+#ifdef USE_EVENT_FD
```
thinkingfish, line 103: shall we model this function, style wise, after the read notification one, and simply include the code from the two write functions (eventfd, pipe) into this one? I think the LOC total is comparable.

src/core/data/worker.c
```diff
@@ -93,12 +93,12 @@ static void
 worker_add_stream(void)
```
thinkingfish, line 93: Do you mind fixing the naming inconsistency _I_ introduced? we can call thish one `_worker_read_notification`, and the one used for return `_worker_write_notification`

src/core/data/worker.c
```diff
@@ -153,6 +175,17 @@ _worker_pipe_write(void)
         log_error("could not write to pipe - %s", strerror(pipe_term->err));
     }
 }
+#endif
+
+static inline void
+_worker_notify_server(void)
```
thinkingfish, line 182: See previous comment, rename to `_worker_write_notification`

src/core/data/worker.c
```diff
@@ -153,6 +175,17 @@ _worker_pipe_write(void)
...
+{
+#ifdef USE_EVENT_FD
+    _worker_write_event_fd();
```
thinkingfish, line 185: And similar to comment to the server side code, maybe just expand everything in this one function.

src/core/data/server.c
thinkingfish: I meant `write` and `read`

PR #295, opened 2020-08-06
add flush to memcached to allow prefix-based, forced key expiration

thinkingfish:


src/storage/slab/item.c
```diff
@@ -391,3 +391,40 @@ item_flush(void)
...
+        struct item_slh *entry = &hash_table->table[i];
+        struct item *it;
+
+        SLIST_FOREACH(it, entry, i_sle) {
+            klen = it->klen;
+            vlen = it->vlen;
+            if (klen >= prefix->len &&
+                    cc_bcmp(prefix->data, item_key(it), prefix->len) == 0) {
+                nkey++;
+                it->expire_at = time_proc_sec();
```
noxiouz, line 415: I guess it might worth caching time_proc_sec value

thinkingfish: `time_proc_sec` is inlined and just a load: https://github.com/twitter/pelikan/blob/master/src/time/time.h#L121

PR #300, opened 2020-11-17
add TLS support to rust pingserver

brayniac:
Adds TLS support to the rust pingserver using rustls. This allows
us to support TLS 1.3 and 1.2. TLS handshake occurs on the server
thread, so new connections will not interfere with request
processing on the worker thread.

Adds a simple pingclient with TLS support so we can get basic perf
numbers to compare with and without TLS.

Fixes an issue where the rust pingserver may fail to accept new
connections.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -77,6 +77,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "base64"
+version = "0.12.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3441f0f7b02788e948e47f457ca01f1d7e6d92c693bc132c22b087d3141c03ff"
+
+[[package]]
 name = "bindgen"
 version = "0.53.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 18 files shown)


src/rust/config/src/server.rs
```diff
@@ -7,7 +7,7 @@ use serde::{Deserialize, Serialize};
 use std::net::{AddrParseError, SocketAddr};
 
 // constants to define default values
-const SERVER_HOST: &str = "0.0.0.0";
+const SERVER_HOST: &str = "127.0.0.1";
```
brayniac, line 10: This shouldn't have changed

src/client/pingclient-rs/src/buffer.rs
```diff
@@ -0,0 +1,327 @@
+// Copyright 2019 Twitter, Inc.
```
brayniac, line 1: This whole file is copy-pasted from the pingserver - we should factor out into a separate buffer library - maybe in rustcommon?

thinkingfish (approved):

PR #309, opened 2020-12-04
new integration test for rust pingserver

brayniac:
Adds a new integration test for the rust pingserver driven directly
by `cargo test`. This improves the developer experience while
working on the rust pingserver by making the test experience more
familiar.


brayniac:
See https://doc.rust-lang.org/book/ch11-03-test-organization.html#integration-tests-for-binary-crates for why this required converting the pingserver to a library+binary crate.

src/server/pingserver-rs/benches/benchmark.rs
```diff
@@ -0,0 +1,45 @@
...
+use pelikan_pingserver_rs::PingserverBuilder;
+
+use std::io::{Read, Write};
+use std::net::TcpStream;
+use std::time::Duration;
+
+fn pingserver_benchmark(c: &mut Criterion) {
+    // launch the pingserver and wait until its ready
+    let pingserver = PingserverBuilder::new(None).spawn();
+    std::thread::sleep(Duration::from_millis(100));
```
thinkingfish, line 15: Is there a chance that this step is flaky on a very slow, shared testing platform?

src/server/pingserver-rs/src/worker.rs
```diff
@@ -73,74 +74,81 @@ impl Worker {
...
+                if event.is_readable() {
+                    self.increment_count(&Stat::WorkerEventRead);
+                    let _ = self.do_read(token);
+                }
 
-                    if event.is_writable() {
-                        self.increment_count(&Stat::WorkerEventWrite);
-                        self.do_write(token);
-                    }
+                if event.is_writable() {
```
thinkingfish, line 90: now that I think about it, maybe we should handle write event before read event if they are both present. The existing C logic has the same order but that's not optimal. Since write event only occurs when the last write was blocked, this allows us to flush the write buffer before handling (potential) new reads, and therefore reduces the chance of having to expand buffer due to new response on top of residual data.

src/server/pingserver-rs/src/worker.rs
```diff
@@ -73,74 +74,81 @@ impl Worker {
...
-                            }
-                        };
+                    Err(_) => {
+                        error!("Error registering new socket");
+                    }
+                };
+            }
+
+            // poll queue to receive new messages
+            while let Ok(message) = self.message_receiver.try_recv() {
```
thinkingfish, line 136: handle messages before sessions for better request latency?

thinkingfish (commented):
LG, only a few comments.

thinkingfish:
> See https://doc.rust-lang.org/book/ch11-03-test-organization.html#integration-tests-for-binary-crates for why this required converting the pingserver to a library+binary crate.

Having the binary as a library is a useful idea beyond testing. I can see this being handy in all kinds of embedding.

src/server/pingserver-rs/benches/benchmark.rs
brayniac: Yeah - I'll increase that - a few seconds should be plenty for the server to startup, 100ms could wind up as flakey in the future.

src/server/pingserver-rs/src/worker.rs
brayniac: Maybe?

The session may already have pending data to read - which we are handling immediately after registering the new session to the worker loop. I don't think I'd want to delay that processing until after the message queue is checked?

Definitely possible that data has arrived since that is not in the userspace buffer yet. But if the client hasn't already sent, it's possible it's just an idle connection in a pool? Whereas if we handle the messages before new sessions, we are definitely delaying the response for an eager client that sent a request already.

We should probably add some stats around this, would be an interesting thing to track.

Maybe a compromise is to set a bound on the number of messages we try to handle here? Since the only message is the shutdown message... it doesn't matter much currently. Any message will cause the thread to terminate.

brayniac:
> > See https://doc.rust-lang.org/book/ch11-03-test-organization.html#integration-tests-for-binary-crates for why this required converting the pingserver to a library+binary crate.
> 
> Having the binary as a library is a useful idea beyond testing. I can see this being handy in all kinds of embedding.

Yeah - the more I've thought about it, the more I like it as a pattern =)

src/server/pingserver-rs/src/worker.rs
brayniac: Reordered for admin/server/worker for consistency. Comments added to indicate reasoning for the order.

src/server/pingserver-rs/src/worker.rs
thinkingfish: If a read event occurs on an existing socket, chances are that client is also “eager”, unless it’s an EOF/FIN which is less likely. Here the consideration is more about minimizing connection churn. By prioritizing (events on) fully established connections, we aim to reduce the chance of client timeouts on these connections and limit the churn to a smaller, more predictable subset of connections. 

Memcached has a fairness control mechanism to limit the number of processed requests per connection per event, we can reintroduce it if pipelining is more widely observed (eg if proxy is used commonly). But that works at a different level.

brayniac: So, in this design - we use the 'server' thread to drive the TCP listener - calling accept as necessary and driving any TLS handshake. By the time the connection is available on the queue for the worker to take ownership and register with its epoll, it's already fully established.

There are three cases related to this:
1 - worker receives a new fully established connection, the client has not sent data and will not for some time, the connection is registered into the 'worker' epoll loop and receives an event at some point in the future.
2 - worker receives a new fully established connection, this connection already has application data in the receive buffer (TLS early data, 0-rtt). This case is handled at https://github.com/twitter/pelikan/pull/309/files#diff-c1ae6312de0ac5c3c9fd29b44cd4d3304be7e5dbd66aceff207fea90d4045cf2R131 - we would not receive another readable event. 
3 - worker receives a new fully established connection, this connection has no application data in the receive buffer yet, but would not block if read were called immediately. This would be the result of a client sending application data while the 'worker' thread is handling other requests.

What I meant in my previous comment was that if we reorder message handling before session handling, we force that delay onto any early data request (assuming the worker is spending most of its time handling requests or waiting for poll to timeout).

thinkingfish (approved):

PR #310, opened 2021-01-04
pingserver-rs: use strum to derive metric names and enum iterator

brayniac:
By using strum derive macros, we are able to simplify the metrics
definitions. As our metric names are well-formed enum variants, we
can simply convert the variant name to snake case and make use of
the enum variant iterator to handle registration.


thinkingfish (approved):
Nice/Terse way of expanding.

PR #311, opened 2021-01-04
pingserver-rs: fix clippy lints

brayniac:
Address some clippy lints and add explicit allows where the
exception improves performance or clarity of the code.


brayniac:
These lints are fixed in #313 - can close this w/o merge after that is landed

brayniac:
This was fixed by #313. Closing without merge

PR #313, opened 2021-01-25
pingserver-rs: replace rustls with boringssl

brayniac:
Replaces rustls with boringssl. This will be a better dependency
for long-term at the cost of an initial performance regression.

### DIFF
--- a/src/server/pingserver-rs/src/admin.rs
+++ b/src/server/pingserver-rs/src/admin.rs
@@ -106,61 +106,39 @@ impl Admin {
                 if event.token() == Token(LISTENER_TOKEN) {
                     while let Ok((stream, addr)) = self.listener.accept() {
                         // handle TLS if it is configured
-                        if let Some(ssl_context) = &self.ssl_context {
+                        let mut session = if let Some(ssl_context) = &self.ssl_context {
                             match Ssl::new(&ssl_context).map(|v| v.accept(stream)) {
                                 // handle case where we have a fully-negotiated
                                 // TLS stream on accept()
                                 Ok(Ok(tls_stream)) => {
-                                    let mut session =
-                                        Session::tls(addr, tls_stream, self.metrics.clone());
-                                    trace!("accepted new session: {}", addr);
-                                    let s = self.sessions.vacant_entry();
-                                    let token = s.key();
-                                    session.set_token(Token(token));
-                                    if session.register(&self.poll).is_ok() {
-                                        s.insert(session);
-                                    } else {
-                                        let _ =
-                                            self.metrics().increment_counter(&Stat::TcpAcceptEx, 1);
-                                    }
+                                    Session::tls(addr, tls_stream, self.metrics.clone())
                                 }
-                                // handle case where further negotiation is
-                                // needed
                                 Ok(Err(HandshakeError::WouldBlock(tls_stream))) => {
-                                    let mut session = Session::handshaking(
+                                    Session::handshaking(
                                         addr,
                                         tls_stream,
                                         self.metrics.clone(),
-                                    );
-                                    let s = self.sessions.vacant_entry();
-                                    let token = s.key();
-                                    session.set_token(Token(token));
-                                    if session.register(&self.poll).is_ok() {
-                                        s.insert(session);
-                                    } else {
-                                        let _ =
-                                            self.metrics().increment_counter(&Stat::TcpAcceptEx, 1);
-                                    }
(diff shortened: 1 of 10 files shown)


brayniac:
pingserver-rs_ssl_comparison

Note on interpreting the chart, the X axis the the request ratelimit in rpc-perf. It does not mean that the backend successfully hit that throughput. Additionally, the line is a centered moving average with a window of 10 points. Points are evenly spaced at 500qps intervals. All testing done over loopback (lo0) with 25 client threads and a total of 2500 connections to the pingserver. Test was run for 1 minute and reports the latency across the first minute of the test.

Attaching latency profile of rustls, openssl, and boringssl variants of the pingserver-rs backend. While boringssl is a slight reduction in throughput using a p999 < 5ms metric and in peak throughput, this regression appears to be an acceptable trade-off in taking a healthier dependency.

Peak Throughput:
rustls: 76500 qps
boring: 74000 qps
openssl: 74000 qps

p999 < 5ms:
rustls: ~69000 qps
boring: ~68000 qps
openssl: ~16000 qps

src/server/pingserver-rs/src/common.rs
```diff
@@ -10,78 +11,26 @@ pub enum Message {
...
-                |_| {
-                    std::io::Error::new(
-                        std::io::ErrorKind::Other,
-                        "Could not parse certificate file",
-                    )
-                },
-            )?,
-        )
+    if let Some(f) = config.tls().certificate_chain() {
+        builder.set_ca_file(f)?;
```
brayniac, line 18: Error handling can be improved here when file doesn't exist.

src/server/pingserver-rs/src/admin.rs
```diff
@@ -104,33 +105,61 @@ impl Admin {
...
+                                    );
+                                    let s = self.sessions.vacant_entry();
+                                    let token = s.key();
+                                    session.set_token(Token(token));
+                                    if session.register(&self.poll).is_ok() {
+                                        s.insert(session);
+                                    } else {
+                                        let _ =
+                                            self.metrics().increment_counter(&Stat::TcpAcceptEx, 1);
+                                    }
```
thinkingfish: line 135-143 is a duplicate of 117-125? Can we shuffle the order of these cases and use fall through to avoid duplication?

src/server/pingserver-rs/src/admin.rs
```diff
@@ -104,33 +105,61 @@ impl Admin {
...
-                                stream,
-                                State::Established,
-                                None,
-                                self.metrics.clone(),
-                            );
-                            trace!("accepted new admin session: {}", addr);
+                            // handle plain sessions
+                            let mut session = Session::plain(addr, stream, self.metrics.clone());
+                            trace!("accepted new session: {}", addr);
                             let s = self.sessions.vacant_entry();
```
thinkingfish: oh more duplication here. I'd say consider if we can lift this logic outside of the `if...else`

src/server/pingserver-rs/src/admin.rs
```diff
@@ -18,7 +19,7 @@ pub struct Admin {
     config: Arc<PingserverConfig>,
     listener: TcpListener,
     poll: Poll,
-    tls_config: Option<Arc<rustls::ServerConfig>>,
+    ssl_context: Option<SslContext>,
     sessions: Slab<Session>,
     metrics: Arc<Metrics<AtomicU64, AtomicU64>>,
```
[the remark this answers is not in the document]
brayniac: Yes! This will be a separate PR - I have a few thoughts on a macro system for metrics that should give us some options here. I'll open a tracking issue.

src/server/pingserver-rs/src/admin.rs
brayniac: Fixing...

src/server/pingserver-rs/src/admin.rs
brayniac: Yeah... feeling like I missed a couple here in the admin. =)

thinkingfish (approved):

src/server/pingserver-rs/src/admin.rs
brayniac: nevermind - this is already fixed =)

brayniac:
Leaving two comments open, as they're tracked to be resolved in #315 and #316

Will proceed to merge and follow-up on those open items in future PRs.

PR #318, opened 2021-02-01
pingserver-rs: bugfix for stats output

### LINKED ISSUE #317: pingserver-rs: improperly terminated stats response
Admin command `stats` doesn't terminate with `END\r\n` as expected.

Discovered while working on #312 

Fix underlying issue and add some tests.

brayniac:
Fixes a bug in the stats output corresponding to issue #317 by
adding the terminating "END\r\n" to the admin response.

Fixes #317

### DIFF
--- a/src/server/pingserver-rs/tests/integration.rs
+++ b/src/server/pingserver-rs/tests/integration.rs
@@ -13,6 +13,12 @@ use std::io::{Read, Write};
 use std::net::TcpStream;
 use std::time::Duration;
 
+pub const ADMIN_ENDPOINT: &str = "127.0.0.1:9999";
+pub const SERVER_ENDPOINT: &str = "127.0.0.1:12321";
+
+pub const STARTUP_DELAY: Duration = Duration::from_secs(10);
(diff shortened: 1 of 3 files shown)


src/server/pingserver-rs/tests/integration.rs
```diff
@@ -26,7 +32,7 @@ fn main() {
 
     // wait for server to startup. duration is chosen to be longer than we'd
     // expect startup to take in a slow ci environment.
-    std::thread::sleep(Duration::from_secs(10));
+    std::thread::sleep(STARTUP_DELAY);
```
thinkingfish, line 35: I wonder if we should eventually move to polling a "ready-to-serve" flag/variable.

thinkingfish (approved):
LGTM

src/server/pingserver-rs/tests/integration.rs
brayniac: Yeah - this delay definitely feels long during development. Will think about how to achieve this - we'll likely have common testing functionality as more servers are implemented.

Issue #314, opened 2021-01-28
pingserver-rs: add rusage metrics


brayniac:
Add rusage metrics to pingserver-rs

brayniac:
This was closed by #319

PR #319, opened 2021-02-01
pingserver-rs: add rusage metrics

brayniac:
Adds rusage metrics to pingserver to address #314

Replaces the current metrics library with one being developed for
twemcache-rs. It removes overhead of a hash lookup for each metric
increment by using a fixed indexing from the metric enum into a
vec of channels.

On a test host this is giving ~4-5% gain in peak throughput.

### DIFF
--- a/src/server/pingserver-rs/src/admin.rs
+++ b/src/server/pingserver-rs/src/admin.rs
@@ -9,10 +9,8 @@ use crate::*;
 use boring::ssl::{HandshakeError, Ssl, SslContext};
 
 use mio::net::TcpListener;
-use strum::IntoEnumIterator;
-
-use std::convert::TryInto;
 use std::io::{BufRead, ErrorKind};
+use strum::IntoEnumIterator;
 
 /// A `Admin` is used to bind to a given socket address and handle out-of-band
 /// admin requests.
@@ -23,7 +21,6 @@ pub struct Admin {
     poll: Poll,
     ssl_context: Option<SslContext>,
     sessions: Slab<Session>,
-    metrics: Arc<Metrics<Stat>>,
     message_receiver: Receiver<Message>,
     message_sender: SyncSender<Message>,
 }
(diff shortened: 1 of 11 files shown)


src/server/pingserver-rs/src/admin.rs
```diff
@@ -200,13 +189,61 @@ impl Admin {
...
+            ru_msgrcv: 0,
+            ru_nsignals: 0,
+            ru_nvcsw: 0,
+            ru_nivcsw: 0,
+        };
+
+        if unsafe { libc::getrusage(libc::RUSAGE_SELF, &mut rusage) } == 0 {
+            set_counter!(
+                &Stat::RuUtime,
+                rusage.ru_utime.tv_sec as u64 * 1000000000 + rusage.ru_utime.tv_usec as u64 * 1000,
```
thinkingfish, line 222: I think it's more readable if we create and use `fn timeval_to_ns()`

src/server/pingserver-rs/src/admin.rs
```diff
@@ -228,35 +265,42 @@ impl EventLoop for Admin {
...
+                                    Source::Gauge => {
+                                        data.push(format!(
+                                            "STAT {} {}\r\n",
+                                            metric,
+                                            get_gauge!(&metric).unwrap_or(0)
+                                        ));
+                                    }
+                                    Source::Counter => {
+                                        data.push(format!(
+                                            "STAT {} {}\r\n",
```
thinkingfish, line 282: more readable with `fn format_stat()` or something like that.

src/server/pingserver-rs/src/session.rs
```diff
@@ -126,14 +110,12 @@ impl Session {
         match read_result {
             Ok(Some(0)) => Ok(Some(0)),
             Ok(Some(bytes)) => {
-                let _ = self
-                    .metrics
-                    .increment_counter(&Stat::TcpRecvByte, bytes.try_into().unwrap());
+                increment_counter_by!(&Stat::TcpRecvByte, bytes as u64);
```
thinkingfish, line 113: Do we need to track tls read/write aside from tcp read/write? Or only one or the other depending on the socket type?

brayniac: I think we should track both if we have TLS enabled. Some of this will change in an upcoming PR. I will be adding a thin wrapper around TcpStream that ensures all read/write stats are tracked there - will reduce need to increment independently in admin/worker

brayniac: Created #320 to track this

thinkingfish (approved):

PR #322, opened 2021-03-01
Add Segcache experimental code 

1a1a11a:
Most of the code has been cleaned up, but there are still some crypto-like code, I am happy to change them. 
The hash table code has a lot of macros, I can change to inline func if that is preferred. 
There is a concurrency bug after I clean the code, still searching for it, will update it when I find it, but it should not affect the overall structure.


.github/workflows/ci.yml
```diff
@@ -1,8 +1,8 @@
 on:
   push:
-    branches: [ master ]
+    branches: [ master, SegcacheExp ]
```
thinkingfish, line 3: Can we clean this up?

thinkingfish (commented):
If I merge the latest ccommon before your test code would it break you subsequent PR?

.github/workflows/ci.yml
1a1a11a: No, it will not break my code. 
I will clean this up in the last commit, so that I can use the test to make sure I do not break anything. 

> If I merge the latest ccommon before your test code would it break you subsequent PR?

thinkingfish (approved):

PR #334, opened 2021-07-28
protocol: improve memcache parse performance

brayniac:
Improves memcache request parsing performance by replacing the
use of the pair of slice windows with a manual implementation
which will find specific byte sequences.

This results in better performance by reducing the need to scan
through a very long buffer to find matches for tokenization, as
we can just take the first match as check if it is the sequence
we were looking for or if we need to continue reading.

Adds benchmarks which show an improvement with this change.


thinkingfish:
What's old is new again 😉  It seems well worth it.

thinkingfish (approved):

PR #335, opened 2021-08-19
logger: add klog functionality for segcache-rs

brayniac:
Adds a new logger which has command logger specific functionality.

Adds command logging to pelikan_segcache_rs.

Changes to the memcache wire protocol to support providing the
necessary context for command logging.


src/rust/protocol/src/memcache/wire/response/mod.rs
```diff
@@ -19,53 +26,259 @@ pub enum MemcacheResponse {
...
+        }
+    }
+
+    fn code(&self) -> usize {
+        match self {
+            Self::Deleted => 7,
+            Self::Exists => 6,
+            Self::NotFound => 8,
+            Self::NotStored => 9,
+            Self::Stored => 9,
```
thinkingfish, line 51: this should be 5?

thinkingfish: Since this is done for backward compatibility, maybe add a reference to https://github.com/twitter/pelikan/blob/master/src/protocol/data/memcache/response.h

src/rust/config/src/units.rs
```diff
@@ -0,0 +1,4 @@
+pub const B: usize = 1;
```
brayniac, line 1: Missing copyright header

src/rust/logger/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.1.0"
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+chrono = "0.4.19"
+config = { path = "../config" }
+crossbeam = "*"
+log = { version = "0.4.8", features = ["std", "release_max_level_debug"] }
```
brayniac, line 12: Should we remove this to allow 'trace' level for 'release' builds?

src/rust/storage/seg/src/hashtable/mod.rs
```diff
@@ -511,10 +511,6 @@ impl HashTable {
 
         trace!("hash: {} mask: {} bucket: {}", hash, self.mask, bucket_id);
 
-        if cas != get_cas(bucket.data[0]) {
```
brayniac, line 514: Add some tests for CAS

brayniac: Moved this into #337

src/rust/config/src/klog.rs
```diff
@@ -0,0 +1,123 @@
+// Copyright 2020 Twitter, Inc.
```
brayniac, line 1: Year is wrong

src/rust/logger/src/backend/file.rs
```diff
@@ -0,0 +1,252 @@
...
+}
+
+impl FileLogReceiver {
+    /// Flush should be periodically called to write log messages to the output
+    /// and rotate the log file if it crosses the size threshold.
+    pub fn flush(&mut self) {
+        while let Some(mut msg) = self.receiver.pop() {
+            let _ = self.writer.write(&msg);
+            // recycle the buffer if it's not oversized
+            if msg.capacity() <= self.msg_size {
```
brayniac, line 60: It might be better to shrink it and put it back into the pool

src/rust/logger/src/backend/file.rs
```diff
@@ -0,0 +1,252 @@
...
+        self.backup_path = path;
+        self
+    }
+
+    /// Consume the builder and return a configured `FileLogSender` and
+    /// `FileLogReceiver`.
+    pub fn build(self) -> (FileLogSender, FileLogReceiver) {
+        let msg_count = self.buf_size / self.msg_size;
+        let log_queue = Queue::with_capacity(msg_count);
+        let buf_queue = Queue::with_capacity(msg_count);
```
brayniac, line 196: Preallocate the buffers?

src/rust/logger/src/format.rs
```diff
@@ -0,0 +1,23 @@
+use crate::Record;
+use rustcommon_time::{DateTime, Local, SecondsFormat};
+
+pub type FormatFunction = fn(
+    write: &mut dyn std::io::Write,
+    now: DateTime<Local>,
+    record: &Record,
+) -> Result<(), std::io::Error>;
+
+pub fn default_format(
```
brayniac, line 10: Need a klog format too

src/rust/logger/src/builder.rs
```diff
@@ -0,0 +1,104 @@
...
+                    .map(|f| Path::new(f).to_owned()),
+            )
+            .max_size(self.debug.log_max_size())
+            .build();
+
+        let (klog_send, klog_recv) = if let Some(_file) = self.command.file() {
+            let (s, r) = FileLogBuilder::default()
+                .msg_size(DEFAULT_MSG_SIZE)
+                .buf_size(self.command.nbuf())
+                .format(default_format)
```
brayniac, line 62: should be klog_format once implemented

src/rust/logger/src/backend/file.rs
brayniac: Changed the logic to match this

src/rust/logger/src/backend/file/sender.rs
```diff
@@ -0,0 +1,51 @@
+// Copyright 2021 Twitter, Inc.
+// Licensed under the Apache License, Version 2.0
+// http://www.apache.org/licenses/LICENSE-2.0
+
+use crate::*;
+
+/// A `FileLogSender` implements `Log` and can be registered as the main logger.
```
thinkingfish, line 7: To me the destination of the log messages (a file, a Kafka producer, another socket...) should probably only matter to the receiver, which uses that info to carry out the `flush` operation. While there may be some implications to the sender (e.g. we may want to serialize things differently based on destination), characterizing senders directly on that seems like leaky abstraction.

src/rust/logger/src/backend/file/sender.rs
```diff
@@ -0,0 +1,51 @@
...
+/// At the logging callsite, messages will be filtered and formatted before
+/// being sent over a queue to the corresponding `FileLogReceiver`. To reduce
+/// runtime allocations, a pool of buffers is maintained and recycled for
+/// re-use.
+pub struct FileLogSender {
+    // level filter for determining if a log message should be logged
+    pub(crate) level_filter: LevelFilter,
+    // a queue for submitting log messages to the receiver
+    pub(crate) sender: Queue<Vec<u8>>,
```
thinkingfish, line 16: `sender`->`log_queued`
`buf_pool`->`log_cleared`?

src/rust/logger/src/backend/file/sender.rs
thinkingfish: Maybe we can have `PlainTextSender` `BinarySender` etc

brayniac: This is fair. Good suggestion for cleaning that up. Can probably make the Sampling* variant generic across some type implementing Log.

src/rust/logger/src/backend/file/sender.rs
brayniac: `log_tx`, `buf_rx`? would be fine with your suggestions here otherwise

thinkingfish: I generally want to avoid things like `tx` and `rx` that very much depends on the point of view, unless it's very obvious (e.g. in a server or client). A sender's tx is a receiver's rx, for example. Instead we can maybe focus on the state of these log containers.

But I'm not super opinionated on exactly how to name these, e.g. I'm equally fine with `log_filled` instead of `log_queued` and `log_free` instead of `log_cleared`.

thinkingfish:
For logger in general, I wonder if we can convey the notion that the `logger` as instantiated in this project is really more of an aggregate of multiple logging lanes. (Maybe a `LogBus` ?)

I say this because it's not super common for a program to have multiple loggers of different nature coexist under a single entity. Most applications have debug logs, Finagle also has traces. In C-Pelikan, we hooked in `dlog` and `klog` separately at the top level in main. So it's easy for people to overlook the fact that both debug and command logger now are delivered by the same struct unless they peek into the logger module itself.

src/rust/logger/src/receiver.rs
```diff
@@ -0,0 +1,26 @@
...
+// http://www.apache.org/licenses/LICENSE-2.0
+
+use crate::*;
+
+/// A `LogReceiver` is the receiving end of the log queues. It should be owned
+/// by the thread which will be flushing the log messages to their respective
+/// outputs. It is expected that the `flush()` function will be periodically
+/// called to make room for new log messages.
+pub struct LogReceiver {
+    pub(crate) debug: FileLogReceiver,
```
thinkingfish, line 12: I think it's OK to make debug logging optional too. We normally don't do that but it can be used to see if logging introduces any meaningful overhead. When I first put logging into Pelikan-C I saw it popping up in profiling, until I changed the macro implementation to move level comparison logic outside the function body.

src/rust/logger/src/sender.rs
```diff
@@ -0,0 +1,40 @@
+// Copyright 2021 Twitter, Inc.
+// Licensed under the Apache License, Version 2.0
+// http://www.apache.org/licenses/LICENSE-2.0
+
+use crate::*;
+
+#[derive(Clone)]
+pub struct LogSender {
+    pub(crate) debug: Arc<dyn Log>,
```
thinkingfish, line 9: The thing that bugs me slightly about this struct is that level is only applicable to debug, and command needs a different set of options (for now, sampling rate). Can we bundle these sender specific configs with the corresponding Log instead of listing them side by side?

src/rust/logger/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+
+////////////////////////////////////////////////////////////////////////////////
+// TODO(bmartin): everything below is Pelikan specific, and should be factored
+// out into a helper when we move this crate into rustcommon
+////////////////////////////////////////////////////////////////////////////////
+
+#[macro_export]
+macro_rules! klog {
+    ($($arg:tt)*) => (
+        error!(target: "klog", $($arg)*);
```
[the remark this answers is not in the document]
brayniac: Not strictly true - there's a `LevelFilter::Off` variant exists, which will filter `Level::Error`, but added a comment that conveys what you're asking for. =)

src/rust/logger/src/format.rs
```diff
@@ -0,0 +1,41 @@
...
+}
+
+pub fn klog_format(
+    w: &mut dyn std::io::Write,
+    now: DateTime<Local>,
+    record: &Record,
+) -> Result<(), std::io::Error> {
+    writeln!(
+        w,
+        "{} {}",
```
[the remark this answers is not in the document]
brayniac: 👍  agree. #341 opened to track that

src/rust/logger/src/single.rs
```diff
@@ -0,0 +1,178 @@
...
+        }
+        self.output.flush()
+    }
+}
+
+/// A type to construct a basic `AsyncLog` which routes all log messages to a
+/// single `Output`.
+pub struct LogBuilder {
+    total_buffer_size: usize,
+    log_message_size: usize,
```
thinkingfish, line 90: Given the structure of log buffer, would it be better to expose `log_message_size` (or named more explicitly as `single_message_size`) and `log_queue_depth` instead of doing the division based on size?

I think the memory consideration of log buffer has been too small to matter in production, so providing a number-of-message config knob may be more user-friendly.

thinkingfish (approved):

PR #338, opened 2021-09-07
Replace `rustcommon-fastmetrics` with `rustcommon-metrics-v2`

swlynch99:
This PR replaces the `rustcommon-fastmetrics` crate with the new one being designed in `rustcommon-metrics-v2`. This has a couple of advantages:
1. Creating a new metric now only requires declaring static variable with the appropriate attribute.
2. The type of each metric is decided by its actual type as opposed to runtime assertions.
3. Only metrics that are declared in the crates actually added to the final binary are included.

Here's the details involved in actually working with metrics in pelikan now:
- Declaring a new metric involves declaring statics (with optional initializers) using the `pelikan_metrics!` macro. This will declare a static of the same name and declare the metric name to be the lowercased version of said metric name.
- In order to lowercase the names I had to define a proc-macro that does this. Once `std::str::from_utf8_unchecked` becomes stable as const this could be replaced with some const code generated using the macro.
- I've also create a macro that defines a test to ensure that no two metrics have duplicated names. This should be added to any crate that defines metrics to ensure that there aren't duplicate metrics.


src/rust/macros/src/lib.rs
```diff
@@ -0,0 +1,19 @@
+use proc_macro::TokenStream;
```
brayniac, line 1: license header

src/rust/macros/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "macros"
+version = "0.1.0"
+authors = ["Sean Lynch <seanl@twitter.com>"]
+edition = "2018"
+license = "Apache-2.0"
```
brayniac, line 6: probably worth putting a description in there too. I'm wondering if we want to put this as rust/utils or something. But calling it 'macros' for now seems fine.

brayniac (changes requested):

brayniac (approved):
This looks good to me

src/rust/metrics/src/lib.rs
```diff
@@ -2,188 +2,74 @@
...
+///     // calling Counter::new.
+///     static MY_METRIC: Counter;
+/// 
+///     // Creates a gauge metric called "some_other_metric" and initializes 
+///     // it to have the value 8.
+///     pub static SOME_OTHER_METRIC: Gauge = Gauge::with_value(8);
+/// }
+/// ```
+#[macro_export]
+macro_rules! pelikan_metrics {
```
thinkingfish, line 29: I'm again raising the very important (bike-shedding) question on naming 🤪, which is if we want to call the macro `static_metrics` instead of `pelikan_metrics`, given most modules in this project don't explicitly use `pelikan` in their name but generally stick to a functional description.

thinkingfish (commented):
LGTM except for the one Q and the fact that I don't understand how this macro works

src/rust/metrics/src/lib.rs
swlynch99: I think `pelikan_metrics` is somewhat clearer but this isn't that important so I'll just rename it

swlynch99:
Here's my explanation of how the macro works. I'm not sure which part you find confusing so I'll just go through it.

The goal of this macro is to convert
```rust
static SOME_METRIC: Counter;
```
into
```rust
#[metric(name = "some_metric", crate = path::to::rustcommon_metrics)]
static SOME_METRIC: Counter = Counter::new();
```
For the generated `metric` attribute:
- We compute the name using the `to_lowercase!` proc-macro since it's unfortunately not possible to do this using const code as of yet.
- The `crate` parameter is needed since `rustcommon_metrics` isn't going to be available directly in downstream crates so we need to tell the macro where to find it. (this is copied from how `linkme` does the same thing.) 

The last part is the static declaration. This is somewhat more complicated since we'd like to be able to accept both forms below
```rust
static NO_INITIALIZER: Counter;
static WITH_INITIALIZER: Counter = Counter::new(555);
```
and we use the recursive invocation to pick the first expression so we use `$init` if it's passed an `<$ty>::new()` otherwise. The `crate __internal` is just to differentiate between the public API of the macro (declaring statics) and internal helpers.

PR #339, opened 2021-09-10
seg: check last_expired before iterating TtlBuckets

brayniac:
Currently, when expire is called, there is no check of when it was
last called. This will lead to unnecessary iteration across the
TtlBuckets.

Adds a `last_expired` field to the `TtlBuckets` struct and early
returns if expiration would be a no-op.


brayniac:
Offers a slight performance boost under replay traffic (~ 2%) with a multithreaded configuration.

thinkingfish (approved):
LGTM

PR #342, opened 2021-09-28
add server-side latencies to rust servers

brayniac:
Adds server-side request latency distributions to both Rust servers (segcache and pingserver) by adding a timestamp to the session and modifying core.

Need to think through some of the handling challenges for pipelined requests and validate that I've selected the right spots to
timestamp and record latencies.

### DIFF
--- a/src/rust/session/src/lib.rs
+++ b/src/rust/session/src/lib.rs
@@ -10,16 +10,15 @@ mod buffer;
 mod stream;
 mod tcp_stream;
 
-use std::borrow::Borrow;
-use std::borrow::BorrowMut;
-use std::io::BufRead;
-use std::io::{ErrorKind, Read, Write};
+use std::borrow::{Borrow, BorrowMut};
+use std::io::{BufRead, ErrorKind, Read, Write};
 use std::net::SocketAddr;
 
 use boring::ssl::{MidHandshakeSslStream, SslStream};
 use metrics::{static_metrics, Counter};
 use mio::event::Source;
 use mio::{Interest, Poll, Token};
+use rustcommon_time::Instant;
 
 use buffer::Buffer;
 use stream::Stream;
@@ -51,6 +50,7 @@ pub struct Session {
     write_buffer: Buffer,
     min_capacity: usize,
     max_capacity: usize,
+    timestamp: Instant,
 }
 
 impl Session {
@@ -93,6 +93,7 @@ impl Session {
             write_buffer: Buffer::with_capacity(min_capacity),
             min_capacity,
             max_capacity,
+            timestamp: Instant::now(),
         }
     }
 
@@ -181,6 +182,14 @@ impl Session {
 
     pub fn peer_addr(&self) -> Result<SocketAddr, std::io::Error> {
         self.stream.peer_addr()
+    }
+
+    pub fn timestamp(&self) -> Instant {
+        self.timestamp
+    }
+
+    pub fn set_timestamp(&mut self, timestamp: Instant) {
+        self.timestamp = timestamp;
     }
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1078,7 +1078,7 @@ dependencies = [
 [[package]]
 name = "rustcommon-atomics"
 version = "1.0.0"
-source = "git+https://github.com/twitter/rustcommon#ad0674fbc8b1d861a6200a2b17f13831b7f6fbcb"
+source = "git+https://github.com/twitter/rustcommon#f80fb5a83d057179b14e3871f77a5c61175edd75"
 dependencies = [
  "serde",
 ]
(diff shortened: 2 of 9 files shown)


src/rust/entrystore/src/seg/memcache.rs
```diff
@@ -31,25 +35,36 @@ impl MemcacheStorage for Seg {
...
             entry.key(),
             entry.value().unwrap_or(b""),
             Some(&entry.flags().to_be_bytes()),
             ttl,
         ) {
             Ok(_) => Ok(()),
             Err(_) => Err(MemcacheStorageError::NotStored),
+        };
+        let elapsed = now.elapsed();
+        if elapsed > Duration::from_micros(500) {
```
thinkingfish, line 64: I would prefer if the threshold for slow reqeusts is set with other configs related to debug.

src/rust/session/src/lib.rs
```diff
@@ -51,6 +61,19 @@ pub struct Session {
...
+    interest: Interest,
+    // TODO(bmartin): consider moving these fields and associated logic
+    // out into a response tracking struct. It would make the session
+    // type more applicable to clients if we move this out.
+    //
+    /// a timestamp which is used to calculate response latency
+    timestamp: Instant,
+    /// this is a queue of pending response sizes
+    pending_responses: VecDeque<usize>,
+    /// this holds the total response bytes pending so we can know how many
```
thinkingfish, line 74: If I understand correctly, this records the bytes pending _before_ the last call to `compose`, hence we can compare the current write buffer bytes and this value to get the  size of the last response. The comment can be made more clear on how this value is being used.

src/rust/session/src/lib.rs
```diff
@@ -182,6 +222,30 @@ impl Session {
...
+    }
+
+    pub fn finalize_response(&mut self) {
+        let previous = self.pending_bytes;
+        let current = self.write_pending();
+        let len = current - previous;
+
+        self.pending_bytes = current;
+
+        if len == 0 {
```
thinkingfish, line 241: Does this imply the most recent response is empty (e.g. due to `noreply`)? If so, it'd be nice to add a comment with an example, and why we are recording the request duration here.

src/rust/core/server/src/threads/worker/storage.rs
```diff
@@ -58,7 +59,7 @@ where
 
         Ok(Self {
             nevent: config.nevent(),
-            timeout: Duration::from_millis(config.timeout() as u64),
+            timeout: Duration::from_millis(1),
```
thinkingfish, line 62: why was this config taken out?

src/rust/core/server/src/threads/worker/storage.rs
```diff
@@ -88,67 +89,86 @@ where
         loop {
             STORAGE_EVENT_LOOP.increment();
 
+            let now = Instant::now();
             self.storage.expire();
-
-            // get events with timeout
-            if self.poll.poll(&mut events, timeout).is_err() {
-                error!("Error polling");
+            if now.elapsed() > rustcommon_time::Duration::from_micros(500) {
```
thinkingfish, line 94: I don't understand this change.

src/rust/core/server/src/threads/worker/storage.rs
```diff
@@ -88,67 +89,86 @@ where
...
+            let mut worker_pending = self.worker_queues.pending();
+
+            trace!("handling events");
+            let mut empty = false;
+
+            while !empty {
+                empty = true;
+                for id in 0..workers {
+                    if worker_pending[id] > 0 {
+                        if let Ok(message) = self.worker_queues.recv_from(id) {
```
thinkingfish, line 115: If we start breaking down the processing time at different stages and/or on different threads, we should also consider the "outer" duration vs. "inner" duration. The "outer" duration could be the time it takes from writing a request to the (storage) queue to finishing reading the corresponding response; the "inner" duration is what you are measuring here. This will give us a sense of processing latency versus queuing latency. The former indicates a slow step on the storage thread, the latter indicates when we need to consider scale out the storage thread (through partitioning, etc). 

This will probably require a change in the type of content passed through the queue, instead of a simple `Request` object, we will need a `(Request, timestamp)` tuple.

src/rust/core/server/src/threads/worker/storage.rs
brayniac: This will be reverted

src/rust/core/server/src/threads/worker/storage.rs
brayniac: To be reverted.

src/rust/session/src/lib.rs
brayniac: Yes, comment added

PR #343, opened 2021-10-07
seg: cleanup item removal

brayniac:
Looking at the `HashTable::remove_from()` function, it should be
patterned more like `relink_item()`. This allows us to avoid a full
key comparison.

Also addresses some clippy lints.

### DIFF
--- a/src/rust/storage/seg/src/segments/segment.rs
+++ b/src/rust/storage/seg/src/segments/segment.rs
@@ -322,13 +322,6 @@ impl<'a> Segment<'a> {
         self.check_magic();
     }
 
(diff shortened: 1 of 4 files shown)


src/rust/storage/seg/src/segments/segment.rs
```diff
@@ -613,9 +606,10 @@ impl<'a> Segment<'a> {
 
             debug_assert!(item.klen() > 0, "invalid klen: ({})", item.klen());
 
+            items += 1;
```
thinkingfish, line 609: Do we still need line 627, 628 after lifting these two lines from the first branch below?

PR #351, opened 2021-10-14
bugfix: session memory growth

brayniac:
Under long-running tests, memory utilization increased due to
memory being used in the sessions. This was largely due to the
latency tracking added in #342 and some bugs in how the pending
response length was determined for each finalized response.

This bug resulted in unbounded growth of the `VecDeque` that is
used to track the pending response lengths.

This PR addresses two places where memory growth can happen in
the session as well as the root issue with response length tracking.

For the response length / latency tracking, we fix the calculation of
pending response length and explicitly handle all potential cases. In
addition, we replace the `VecDeque` with a fixed-size array to
eliminate runtime allocations for tracking and bound the memory
used for this feature.

Additionally, the session buffer resize logic has been refactored to
be more straightforward and adds tests to check that the actual
behavior matches expectations.

This combination of fixes avoids unbounded RSS growth within the
sessions during long-running tsts

### DIFF
--- a/src/rust/session/src/buffer.rs
+++ b/src/rust/session/src/buffer.rs
@@ -293,7 +293,7 @@ mod tests {
         // consume all but the last byte of content in the buffer, one byte at
         // a time
         // - buffer len decreases with each call to consume()
-        // - buffer available capacity remains the same
+        // - buffer available capacity increases with each call to consume()
         for i in 1..len {
             buffer.consume(1);
             assert_eq!(buffer.len(), len - i);
@@ -342,14 +342,15 @@ mod tests {
         assert_eq!(buffer.available_capacity(), 1);
         assert!(!buffer.is_empty());
 
-        // partial consume, len decrease, capacity remains the same
+        // partial consume, len decrease, buffer shrinks down to target capacity
         // length = 16, size = 16, capacity = 0
         buffer.consume(15);
         assert_eq!(buffer.len(), 16);
         assert_eq!(buffer.available_capacity(), 0);
 
-        // consume one more byte and the buffer shrinks because we have less
-        // than half occupancy
+        // from here on, buffer will not shrink below target capacity
+
+        // consume one more byte
         // length = 15, size = 16, capacity = 0
         buffer.consume(1);
         assert_eq!(buffer.len(), 15);


src/rust/session/src/lib.rs
```diff
@@ -253,8 +260,13 @@ impl Session {
...
             let latency = (now - self.timestamp()).as_nanos() as u64;
             REQUEST_LATENCY.increment(now, latency, 1);
-        } else {
-            self.pending_responses.push_back(len);
+        } else if self.pending_count < self.pending_responses.len() {
+            let mut idx = self.pending_head + self.pending_count;
+            if idx >= self.pending_responses.len() {
+                idx %= self.pending_responses.len();
+            }
+            self.pending_responses[idx] = len;
```
brayniac, line 268: Blocking issue: the stat here is somehow not quite right...

src/rust/session/src/buffer.rs
```diff
@@ -75,35 +79,47 @@ impl Buffer {
...
-                    self.buffer.set_len(self.buffer.capacity());
-                }
+
+            let target = self.buffer.len() / 2;
+
+            self.buffer.truncate(target);
+            self.buffer.shrink_to_fit();
+        }
+
+        // update stats if the buffer has resized
```
thinkingfish, line 109: This seems more verbose than duplicating the `Less` condition in two places above.

src/rust/session/src/lib.rs
```diff
@@ -241,21 +249,44 @@ impl Session {
...
+                // We've finalized a response that is zero-length. This is
+                // expected for empty responses such as when handling memcache
+                // requests which specify `NOREPLY`. Since there are no pending
+                // bytes for a zero-length response, we can determine the
+                // latency now.
+                let now = Instant::now();
+                let latency = (now - self.timestamp()).as_nanos() as u64;
+                REQUEST_LATENCY.increment(now, latency, 1);
+            }
+            Ordering::Less => {
```
thinkingfish, line 277: Why wouldn't this lead to `Ordering:Equal` ? If request handling result in a reduction in shrinkage of the write buffer, seems that'd be hinting at buffer corruption and therefore more severe than a warning?

src/rust/session/src/lib.rs
```diff
@@ -345,24 +377,63 @@ impl Write for Session {
...
+                        // The buffer is completely flushed. We have no more
+                        // pending bytes.
+                        self.pending_bytes = 0;
+                    }
+                    Ordering::Greater => {
+                        // This indicates that the tracking is off. Potentially
+                        // due to a protocol implementation that failed to
+                        // finalize some response. We will simply log a warning
+                        // and zero out the pending bytes.
+                        warn!(
```
thinkingfish, line 424: ditto- seems to indicate some kind of data corruption in the write buffer.

src/rust/session/src/buffer.rs
brayniac: Consolidated some of the logic above, see if this looks better now.

PR #352, opened 2021-10-18
Improve CI by properly caching dependencies 

swlynch99:
This PR adds a few different github workflows that should have all the same functionality as the current CI. The difference, however, is that these will cache intermediate build artifacts so that they can be reused between runs. In total this should bring the time needed to run all CI workflows down to about ~10 minutes when the cache is warm.

I've also included the audit and clippy jobs from `rustcommon`. Audit currently fails so I've disabled it at this time.

## Details
- Add 3 new workflows for cmake, cargo, and fuzzing
- Cargo workflows cache dependencies and installed cargo binaries
- Cmake workflows use ccache
- I've included a bugfix for the compile errors that occurs when building with --all-features


.github/workflows/fuzz.yml
```diff
@@ -0,0 +1,38 @@
...
+  CARGO_TERM_COLOR: always
+  RUST_BACKTRACE: full
+  FUZZ_TIME: 300
+  FUZZ_JOBS: 2
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        os: [ ubuntu-18.04, macos-10.15 ]
```
brayniac, line 15: I don't think we need to fuzz on both macos and ubuntu.

PR #362, opened 2021-11-03
bugfix: rust command log freezes

brayniac:
We've observed that the command log output may freeze under high
load. This happens due to the implementation of the
`SamplingLogger`. By changing the logic, it will be able to
handle the edge case that was unaccounted for in the initial
implementation.

This brings the `SamplingLogger` implementation more in-line
with the original C implementation of klog.

Fixes the `log_skip` metric to represent the number of log messages
omitted due to sampling. Removes `log_skip_bytes` so we don't pay
to format a log message we will omit. Adds `log_drop` and
`log_drop_bytes` to capture log messages which are dropped due to
a full queue between the logger and its drain.


src/rust/logger/src/single.rs
```diff
@@ -73,8 +73,13 @@ pub(crate) struct LogDrain {
 
 impl Drain for LogDrain {
     fn flush(&mut self) -> Result<(), Error> {
+        LOG_FLUSH.increment();
         while let Some(mut log_buffer) = self.log_filled.pop() {
-            let _ = self.output.write(&log_buffer);
+            if let Err(e) = self.output.write_all(&log_buffer) {
+                LOG_FLUSH_EX.increment();
```
thinkingfish, line 79: I think you meant to increment `LOG_WRITE_EX`

thinkingfish (approved):

PR #377, opened 2021-12-06
protocol/memcache: add request cardinality distribution stats

brayniac:
Adds stats to track request cardinality for GET and GETS commands.
This allows us to get insight into batch size distribution over
time which can provide better insight into traffic patterns.

### DIFF
--- a/src/rust/protocol/src/memcache/wire/mod.rs
+++ b/src/rust/protocol/src/memcache/wire/mod.rs
@@ -11,7 +11,7 @@ pub use response::*;
 use super::*;
 use crate::*;
 
-use metrics::{static_metrics, Counter, Relaxed, Heatmap};
+use metrics::{static_metrics, Counter, Heatmap, Relaxed};
 use rustcommon_time::{Duration, Instant};
 
 static_metrics! {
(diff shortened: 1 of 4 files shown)


src/rust/protocol/src/memcache/wire/mod.rs
```diff
@@ -11,15 +11,22 @@ pub use response::*;
 use super::*;
 use crate::*;
 
-use metrics::{static_metrics, Counter};
+use metrics::{static_metrics, Counter, Relaxed, Heatmap};
+use rustcommon_time::{Duration, Instant};
 
 static_metrics! {
     static GET: Counter;
+    static GET_CARDINALITY: Relaxed<Heatmap> = Relaxed::new(||
```
brayniac, line 19: Open questions:
- Should we call this 'cardinality' or 'batch_size'? 
- Do we want to track GET/GETS together or separately?

thinkingfish: I think `gets` will mostly be unbatched, which may skew the results slightly? (`gets` is used mostly for update with `cas`, and batch updates are uncommon). I am OK if we *only* do this for `get` now.

thinkingfish: no opinion on the naming, leaning slightly  toward `cardinality` since it's one word 😄

thinkingfish (approved):

PR #378, opened 2021-12-06
rust/session: track pipeline depth distribution

brayniac:
Add a stat which tracks the depth of the request pipeline per
session. This is achieved by tracking the number of responses
between resets of the session timestamp. This corresponds to the
number of complete requests in the session buffer for a single
read.


thinkingfish (approved):
LGTM

PR #382, opened 2021-12-09
storage/seg: add clear function

brayniac:
Adds a function to clear the stored data.

### DIFF
--- a/src/rust/storage/seg/src/ttl_buckets/ttl_bucket.rs
+++ b/src/rust/storage/seg/src/ttl_buckets/ttl_bucket.rs
@@ -44,6 +44,7 @@ pub struct TtlBucket {
 }
 
 impl TtlBucket {
+    /// Create a new `TtlBucket` which will hold items with the provided TTL.
     pub(super) fn new(ttl: i32) -> Self {
         Self {
             head: None,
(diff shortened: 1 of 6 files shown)


src/rust/storage/seg/src/ttl_buckets/ttl_bucket.rs
```diff
@@ -106,6 +106,34 @@ impl TtlBucket {
         }
     }
 
+    // clear segments from this TtlBucket, returns the number of segments expired
+    pub(super) fn clear(&mut self, hashtable: &mut HashTable, segments: &mut Segments) -> usize {
```
thinkingfish, line 110: I personally would prefer `bulk_clear` over `clear` for all the new `clear` traits, and keep the original `segment.clear` as the only `clear` function.

src/rust/storage/seg/src/ttl_buckets/ttl_buckets.rs
```diff
@@ -125,6 +125,19 @@ impl TtlBuckets {
...
         expired
     }
+
+    pub(crate) fn clear(&mut self, hashtable: &mut HashTable, segments: &mut Segments) -> usize {
+        let start = Instant::now();
+        let mut cleared = 0;
+        for bucket in self.buckets.iter_mut() {
+            cleared += bucket.clear(hashtable, segments);
+        }
+        segments.set_flush_at(CoarseInstant::now());
```
thinkingfish, line 135: Considering the semantics of the memcached `flush_all` command, which allows delayed flushing, I wonder if we should set this outside this function. This way, we can set that `flush_at` in segments when we receive the command, and let the storage module handle the actual flushing internally (later).

src/rust/storage/seg/src/ttl_buckets/ttl_bucket.rs
brayniac: Happy to make the change if you prefer. I was trying to keep the high-level API closer to the std library HashMap: https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.clear

thinkingfish: Ah I see. this is fine

thinkingfish (approved):

PR #384, opened 2021-12-13
rust: add typed values

brayniac:
Adds typed values to the storage layer. This means that we can
represent numeric values such as `u64` and `i64` in a native format
but allow the protocol to present them in a different way. This
also allows the protocol to do the conversion from stringy numbers
to typed values, which can result in space savings and allow for
operations such as increment / decrement in the future.


src/rust/protocol/src/memcache/wire/response/mod.rs
```diff
@@ -251,20 +252,42 @@ impl Compose for MemcacheResponse {
...
+                                    value.len(),
+                                    entry.cas.unwrap_or(0)
+                                )
+                            } else {
+                                write!(dst, " {} {}", entry.flags, value.len())
+                            };
+                            dst.write_all(CRLF.as_bytes());
+                            dst.write_all(value);
+                        }
+                        Value::I64(_) => {}
```
thinkingfish, line 288: not implemented?

brayniac: Per discussion, I'm removing the signed i64 type until we have a concrete use-case but leaving the infrastructure for supporting multiple classes of typed values so we can add more types in the future.

thinkingfish (approved):

PR #385, opened 2021-12-14
rust: fix tls certificate chain loading

brayniac:
Fixes for TLS certificate chain loading when presented with both
a server cert and chain.

### DIFF
--- a/src/rust/common/src/ssl.rs
+++ b/src/rust/common/src/ssl.rs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0
 // http://www.apache.org/licenses/LICENSE-2.0
 
+use boring::x509::X509;
 use boring::ssl::*;
 use config::TlsConfig;
 use std::io::{Error, ErrorKind};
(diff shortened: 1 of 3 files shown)


src/rust/common/src/ssl.rs
```diff
@@ -2,35 +2,71 @@
...
+    if chain && !cert {
+        // assume we have a complete chain: leaf + intermediates + root in one file
+        if let Some(f) = config.certificate_chain() {
+            builder
+                .set_certificate_chain_file(f)
+                .map_err(|_| Error::new(ErrorKind::Other, "bad certificate chain"))?;
+        } else {
+            return Ok(None);
+        }
+    } else if cert && chain {
```
thinkingfish, line 43: nit: `chain && cert` to be as close to the first branch as possible.

src/rust/common/src/ssl.rs
```diff
@@ -2,35 +2,71 @@
...
+    let cert = config.certificate().is_some();
+
+    if chain && !cert {
+        // assume we have a complete chain: leaf + intermediates + root in one file
+        if let Some(f) = config.certificate_chain() {
+            builder
+                .set_certificate_chain_file(f)
+                .map_err(|_| Error::new(ErrorKind::Other, "bad certificate chain"))?;
+        } else {
+            return Ok(None);
```
thinkingfish, line 41: when would this branch be taken (given `chain` is `true`)?

brayniac: Good point. We can just unwrap inside these branches

thinkingfish (approved):

PR #387, opened 2021-12-16
rust/core: allow admin thread to send signals to all threads

brayniac:
Changes how admin initialization errors are handled so that we
get log output if the admin thread fails to initialize.

Changes the relationships for the signal queue so that the admin
thread can now broadcast messages to all threads.

This will enable us to add flush_all functionality in a follow-up
PR.


src/rust/core/server/src/threads/listener.rs
```diff
@@ -38,7 +38,7 @@ pub struct Listener {
     poll: Poll,
     session_queue: QueuePairs<Session, ()>,
     ssl_context: Option<SslContext>,
-    signal_queue: QueuePairs<(), Signal>,
+    signal_queue: QueuePairs<Signal, Signal>,
```
thinkingfish, line 41: In places where we have both directions utilized, it'll be good to add comment on what the Signal is intended (nature of the message) when they are declared, since the signature is symmetrical and abstract.

brayniac: See if the comments in the lastest commit help here or if there's something else you're looking for.

thinkingfish (approved):

PR #391, opened 2022-01-13
docs: readme improvements

brayniac:
Adding some more information at the top of the readme, including
more badges.

Gives Github workflows actual names so we can use them with
shields.io badges for better visual consistency.

### DIFF
--- a/.github/workflows/cargo.yml
+++ b/.github/workflows/cargo.yml
@@ -1,3 +1,5 @@
+name: cargo-build
+
 on:
   push:
   pull_request:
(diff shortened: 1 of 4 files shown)


brayniac:
@thinkingfish - feel free to make changes on this if you have thoughts about wording. The build status badges might not work until after this is merged. If you prefer, I can use the github provided badges in this commit and follow-up with a PR that changes them to shields.io (after the workflows have names) for better visual consistency.

brayniac:
Note: fuzz tests are failing because they don't have lock files, and are therefore using latest rustcommon. Unrelated to this change.

thinkingfish (approved):

PR #392, opened 2022-01-13
rust: move dependencies on rustcommon into one library

brayniac:
Currently, we face some versioning challenges between the Pelikan
workspace and the fuzz testing suite. The fuzz test suite is
excluded from the Pelikan workspace, so does not share the same
lock file. This means that dependencies are not necessarily in
lockstep with the main code.

This change centralizes the dependencies on rustcommon into the
`common` library within the Pelikan workspace. This allows us to
have a single spot to specify an exact revision of rustcommon for
our dependency.

Includes a bit of refactoring so that the `config` crate depends
on `common` and not the other way around.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -241,14 +241,16 @@ name = "common"
 version = "0.1.0"
 dependencies = [
  "boring",
- "config",
+ "rustcommon-metrics-v2",
  "rustcommon-time",
+ "serde",
 ]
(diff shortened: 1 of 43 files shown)


brayniac:
NOTE: this will have merge conflicts with other open PRs

thinkingfish (approved):

PR #383, opened 2021-12-10
WIP: storage/seg: add support for increment and decrement

brayniac:
Adds support for incrementing and decrementing values stored in
the seg storage. This is necessary for implementing the incr/decr
commands in the segcache backend.

PR #395, opened 2022-02-11
rust: add support for increment and decrement

wandaitzuchen:
Picking up where https://github.com/twitter/pelikan/pull/383 left off.
@brayniac @thinkingfish 

---
Notes:
- Display `CLIENT_ERROR` for cases when incrementing/decrementing a key with value type that is not`U64`. 
- Updated tests

Testing:
```
% telnet localhost 12321
...
set key 0 100 3
200
STORED
incr key 5
205
decr key 100
105
incr key2 5
NOT_FOUND
set key 0 100 3
val
STORED
incr key 10
CLIENT_ERROR
```

### DIFF
--- a/src/rust/entrystore/src/seg/memcache.rs
+++ b/src/rust/entrystore/src/seg/memcache.rs
@@ -130,12 +130,22 @@ impl MemcacheStorage for Seg {
         }
     }
 
-    fn incr(&mut self, _key: &[u8], _value: u64) -> Result<u64, MemcacheStorageError> {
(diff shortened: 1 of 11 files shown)


brayniac:
Thanks for opening this. I have a separate branch that I haven't pushed yet which is supposed to address finishing this work. I'll take a look next week and see which we'll want to use to add this functionality.

brayniac:
Closing this as #398 has been merged. Thanks again for the PR!

PR #398, opened 2022-02-15
segcache-rs: add incr/decr support

brayniac:
Adds support for incr/decr commands to segcache-rs.

In the storage library, we explicitly implement these commands
as wrapping_add() and saturating_sub(). This allows us to choose
the behaviors at higher-levels and potentially make additional
operations available in different protocols.

The implementation in segcache-rs aims to be memcached
compatible and returns "ERROR" when numeric operations are
attempted on non-numeric values.

Replaces #383 and #395

### DIFF
--- a/src/rust/entrystore/src/seg/memcache.rs
+++ b/src/rust/entrystore/src/seg/memcache.rs
@@ -132,25 +132,25 @@ impl MemcacheStorage for Seg {
 
     fn incr(&mut self, key: &[u8], value: u64) -> Result<u64, MemcacheStorageError> {
         match self.data.wrapping_add(key, value) {
-            Err(SegError::NotFound) => Err(MemcacheStorageError::NotFound),
-            Err(SegError::NotNumeric) => Err(MemcacheStorageError::Error),
-            Err(_) => Err(MemcacheStorageError::ServerError),
             Ok(item) => match item.value() {
                 Value::U64(v) => Ok(v),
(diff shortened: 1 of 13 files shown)


src/rust/entrystore/src/seg/memcache.rs
```diff
@@ -130,12 +130,28 @@ impl MemcacheStorage for Seg {
...
 
-    fn incr(&mut self, _key: &[u8], _value: u64) -> Result<u64, MemcacheStorageError> {
-        Err(MemcacheStorageError::NotSupported)
+    fn incr(&mut self, key: &[u8], value: u64) -> Result<u64, MemcacheStorageError> {
+        match self.data.wrapping_add(key, value) {
+            Err(SegError::NotFound) => Err(MemcacheStorageError::NotFound),
+            Err(SegError::NotNumeric) => Err(MemcacheStorageError::Error),
+            Err(_) => Err(MemcacheStorageError::ServerError),
+            Ok(item) => match item.value() {
+                Value::U64(v) => Ok(v),
```
thinkingfish, line 139: nitpicking: I wonder if we should either have the `Ok(item) => ... Value::U64(v)` be the first or last line of this entire code bloc, so the error-handling are visually grouped.

brayniac: Good point. That's more consistent with the other blocks too. Fixed and pushed.

thinkingfish (approved):

PR #401, opened 2022-02-22
new queues crate for rust server

brayniac:
The previous queues crate could be a little challenging to work
with and required bookeeping at the callsite to track senders to
return directed responses as well as tracking sent responses to
consolidate calls to wakeup the receiving thread.

The new queues crate attempts to simplify things by tracking if
a wakeup is needed on a per-receiver basis, and provides a
convenience function to receive all pending items in a single
call.

Additionally, instead of doing a predictable round-robin send, we
change the behavior to use a uniform random distribution to pick a
receiver when the communication is from 1 -> any. This helps
provide a better distribution across receivers regardless of the
order in which the items are received.

The core server crate is refactored to have more builder types
which are passed the queues for inter-process communication at time
of finalization. This offers a cleaner way of initializing the
process and connecting the threads with the queues.

Note: this does replace the use of the SPSC queues from `rtrb`
with MPMC queues from `crossbeam-channel`. I do not find any
significant performance penalty to the new implementation in
some casual benchmarking.


src/rust/queues/src/lib.rs
```diff
@@ -2,14 +2,362 @@
...
+    ///
+    /// Since bounded queues are used internally, the capacity is the maximum
+    /// number of pending items each receiver can have. For example, if side A
+    /// has 4 queues, and a capacity of 1024*4 => 4096 items of type U may be
+    /// pending.
+    ///
+    /// NOTE: the return vectors maintain the ordering of the wakers that were
+    /// provided. Care must be taken to ensure that the corresponding queues are
+    /// given to the event loop with the corresponding waker.
+    pub fn new<A: AsRef<[Arc<Waker>]>, B: AsRef<[Arc<Waker>]>>(
```
thinkingfish, line 101: Nice comment/documenation

src/rust/core/server/src/threads/workers/multi.rs
```diff
@@ -186,7 +207,7 @@ where
                 let mut message = TokenWrapper::new(request, token);
 
                 for retry in 0..QUEUE_RETRIES {
-                    if let Err(SendError::Full(m)) = self.storage_queue.try_send(message) {
+                    if let Err(m) = self.storage_queue.try_send_to(0, message) {
```
thinkingfish, line 210: Would  it help readers if storage thread id is declared using a constant instead of `0` explicitly?

thinkingfish (approved):

src/rust/core/server/src/threads/workers/storage.rs
```diff
@@ -0,0 +1,153 @@
...
+}
+
+impl<Storage, Request, Response> StorageWorker<Storage, Request, Response>
+where
+    Storage: Execute<Request, Response> + EntryStore,
+{
+    /// Run the `StorageWorker` in a loop, handling new session events.
+    pub fn run(&mut self) {
+        let mut events = Events::with_capacity(self.nevent);
+        let mut requests = Vec::with_capacity(1024);
```
thinkingfish, line 86: Do you intend to use `QUEUE_CAPACITY` instead of `1024`?

src/rust/queues/src/lib.rs
```diff
@@ -2,14 +2,362 @@
...
+
+    const WAKER_TOKEN: Token = Token(usize::MAX);
+
+    #[test]
+    fn basic() {
+        let poll = Poll::new().expect("failed to create event loop");
+        let waker =
+            Arc::new(Waker::new(poll.registry(), WAKER_TOKEN).expect("failed to create waker"));
+
+        let (mut a, mut b) = Queues::<usize, String>::new(vec![waker.clone()], vec![waker], 1024);
```
thinkingfish, line 284: `QUEUE_CAPACITY`?

src/rust/core/server/src/threads/workers/mod.rs
```diff
@@ -0,0 +1,280 @@
...
+            Self::Multi {
+                storage,
+                mut workers,
+            } => {
+                let storage_wakers = vec![storage.waker()];
+                let worker_wakers: Vec<Arc<Waker>> = workers.iter().map(|v| v.waker()).collect();
+                let (mut response_queues, mut request_queues) =
+                    Queues::new(worker_wakers, storage_wakers, QUEUE_CAPACITY);
+
+                let s = storage.build(signal_queues.remove(0), request_queues.remove(0));
```
thinkingfish, line 200: Can you document the reason to remove the first element?

src/rust/core/server/src/threads/workers/storage.rs
brayniac: Yes, thanks!

src/rust/core/server/src/threads/workers/multi.rs
brayniac: Done

thinkingfish (approved):

PR #410, opened 2022-04-05
segcache optimizations

brayniac:
Some optimizations for segcache performance. On a test system this
gives somewhere around 2% more performance. Benefit may vary
depending on config and workload.

- Reduce calls to Instant::recent() to reduce atomics in hot path.
- Remove trace logging in hot path.
- Change ordering of fields to optimize hot path.
- Use crossbeam_queue instead of crossbeam_channel to remove
  unnecessary channel level wake.

### DIFF
--- a/src/rust/storage/seg/src/hashtable/mod.rs
+++ b/src/rust/storage/seg/src/hashtable/mod.rs
@@ -150,13 +150,13 @@ impl HashTable {
             mask,
             data: data.into_boxed_slice(),
             rng: Box::new(rng()),
-            started: Instant::recent(),
+            started: Instant::now(),
             next_to_chain: buckets as u64,
         }
     }
 
     /// Lookup an item by key and return it
-    pub fn get(&mut self, key: &[u8], segments: &mut Segments) -> Option<Item> {
+    pub fn get(&mut self, key: &[u8], time: Instant, segments: &mut Segments) -> Option<Item> {
         let hash = self.hash(key);
         let tag = tag_from_hash(hash);
         let bucket_id = hash & self.mask;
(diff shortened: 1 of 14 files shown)


src/rust/storage/seg/src/hashtable/mod.rs
```diff
@@ -557,7 +551,7 @@ impl HashTable {
 
                         if cas == get_cas(bucket.data[0]) {
                             // TODO(bmartin): what is expected on overflow of the cas bits?
-                            self.data[(hash & self.mask) as usize].data[0] += 1;
+                            self.data[(hash & self.mask) as usize].data[0] += 1 << 32;
```
thinkingfish, line 554: use `bucket_id as usize` and `CAS_BIT_SHIFT`?

src/rust/storage/seg/src/hashtable/mod.rs
```diff
@@ -488,11 +484,11 @@ impl HashTable {
             insert_item_info = 0;
             self.data[bucket_id].data[N_BUCKET_SLOT - 1] = next_id as u64;
 
-            self.data[(hash & self.mask) as usize].data[0] += 0x0001_0000_0000_0000;
+            self.data[(hash & self.mask) as usize].data[0] += 0x0000_0000_0001_0000;
         }
 
         if insert_item_info == 0 {
-            self.data[(hash & self.mask) as usize].data[0] += 1;
+            self.data[(hash & self.mask) as usize].data[0] += 1 << 32;
```
thinkingfish, line 491: use `CAS_BIT_SHIFT`?

src/rust/storage/seg/src/hashtable/mod.rs
brayniac: Good catch! Fixed

src/rust/storage/seg/src/hashtable/mod.rs
brayniac: Fixed

thinkingfish (approved):

PR #411, opened 2022-04-14
add http admin port to rust servers

brayniac:
Adds a http admin port to the rust servers to expose statistics in
human, json, and prometheus formats.

### DIFF
--- a/config/segcache.toml
+++ b/config/segcache.toml
@@ -6,6 +6,13 @@ host = "0.0.0.0"
 # port listening on
 port = "9999"
 
+# enable the http admin port?
+http_enabled = true
+# http listening interface
+http_host = "0.0.0.0"
+# http listening port
+http_port = "9998"
+
 [server]
 # interfaces listening on
 host = "0.0.0.0"
--- a/src/rust/core/server/src/threads/admin.rs
+++ b/src/rust/core/server/src/threads/admin.rs
@@ -5,7 +5,6 @@
 //! The admin thread, which handles admin requests to return stats, get version
 //! info, etc.
 
-use tiny_http::{Method, Request, Response};
 use crate::poll::{Poll, LISTENER_TOKEN, WAKER_TOKEN};
 use crate::threads::EventLoop;
 use crate::QUEUE_RETRIES;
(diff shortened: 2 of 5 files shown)


config/segcache.toml
```diff
@@ -6,6 +6,13 @@ host = "0.0.0.0"
 # port listening on
 port = "9999"
 
+# enable the http admin port?
```
brayniac, line 9: I considered taking a single `http_listen` parameter and checking for presence or absence to enable/disable the http server, but it's inconsistent with the existing config.

thinkingfish: I can see us moving to default to http, and allowing old style only with a "legacy" option within a year. You can change how config works when we make that transition.

src/rust/core/server/src/threads/admin.rs
```diff
@@ -329,6 +353,147 @@ impl Admin {
         };
     }
 
+    fn human_stats(&self) -> String {
```
thinkingfish, line 356: Add a short description/template to show what human stats format means

src/rust/core/server/src/threads/admin.rs
```diff
@@ -329,6 +353,147 @@ impl Admin {
...
+                    data.push(format!("{}_{}: {}", metric.name(), label, percentile));
+                }
+            }
+        }
+
+        data.sort();
+        data.join("\n")
+    }
+
+    fn json_stats(&self) -> String {
```
thinkingfish, line 383: Add a short description/template to show what json stats format means, since it is conforming to a particular Twitter convention (shall we add a link to the finagle metrics format too?)

src/rust/core/server/src/threads/admin.rs
```diff
@@ -329,6 +353,147 @@ impl Admin {
...
+
+        data.sort();
+        let body = data.join(",");
+        let mut content = head;
+        content += &body;
+        content += "}";
+        content
+    }
+
+    fn prometheus_stats(&self) -> String {
```
thinkingfish, line 416: ditto, a reference would be nice

src/rust/core/server/src/threads/admin.rs
```diff
@@ -329,6 +353,147 @@ impl Admin {
...
+        parts.join("_")
+    }
+
+    /// Handle a HTTP request
+    fn handle_http_request(&self, request: Request) {
+        let url = request.url();
+        let parts: Vec<&str> = url.split('?').collect();
+        let url = parts[0];
+        match url {
+            "/metrics" => match request.method() {
```
thinkingfish, line 467: general comments about endpoint conventions (e.g. wrt prometheus collector) would be helpful

thinkingfish (approved):
Conditional approval given all comments are about adding comments.

PR #413, opened 2022-04-20
segcache: make item header immutable

brayniac:
Makes the item header immutable in the Rust implementation of
Segcache by removing use of the `deleted` bit in the item header.

Since the Rust implementation of segment storage is not concurrent,
there is no need for the `deleted` bit in the item header. Instead,
we can check the hashtable to confirm if the item is located within
a given segment at a specified offset. By removing the use of the
`deleted` bit, we make the item header immutable. This is
beneficial in that it removes the need to perform random updates
within the segments.

### DIFF
--- a/src/rust/storage/seg/src/hashtable/mod.rs
+++ b/src/rust/storage/seg/src/hashtable/mod.rs
@@ -410,6 +410,47 @@ impl HashTable {
         }
     }
 
+    pub(crate) fn is_item_at(&self, key: &[u8], seg: NonZeroU32, offset: u64) -> bool {
+        let hash = self.hash(key);
+        let tag = tag_from_hash(hash);
+        let bucket_id = hash & self.mask;
+
+        let mut bucket = &self.data[bucket_id as usize];
+        let chain_len = chain_len(bucket.data[0]);
+        let mut chain_idx = 0;
+
+        loop {
+            let n_item_slot = if chain_idx == chain_len {
+                N_BUCKET_SLOT
+            } else {
+                N_BUCKET_SLOT - 1
+            };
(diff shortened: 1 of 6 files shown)


src/rust/storage/seg/src/hashtable/mod.rs
```diff
@@ -410,6 +410,47 @@ impl HashTable {
...
+    pub(crate) fn is_item_at(&self, key: &[u8], seg: NonZeroU32, offset: u64) -> bool {
+        let hash = self.hash(key);
+        let tag = tag_from_hash(hash);
+        let bucket_id = hash & self.mask;
+
+        let mut bucket = &self.data[bucket_id as usize];
+        let chain_len = chain_len(bucket.data[0]);
+        let mut chain_idx = 0;
+
+        loop {
```
thinkingfish, line 422: why is this not `for chain_idx in 0..chain_len {` ? It'd be a little more concise that way I think.

brayniac: Excellent question! This should probably be changed in every place we're scanning through the hash buckets. This was just copy past from another function that does something similar. 

It could probably be expressed as some sort of iterator itself... but for now I'll cleanup the loops

brayniac: Fixed.

thinkingfish (approved):

src/rust/storage/seg/src/hashtable/mod.rs
brayniac: #414 for converting to an iterator in the future.

Issue #414, opened 2022-04-22
rust: storage/seg - implement an iterator for item info


brayniac:
In the hashtable module, we are frequently iterating through the hash buckets and info slots to get item info. This is a lot of duplicate code that feels like it can be more concisely expressed at the call sites as an iterator.

hderms:
@brayniac are you referring to iterations like this repeated chunk: 

https://github.com/twitter/pelikan/blob/13b1b3ac1fd3f6181bee790e64133fbd125842b3/src/rust/storage/seg/src/hashtable/mod.rs#L307-L318

brayniac:
@hderms - yes. It feels like we can provide an iterator where the iterator's item type is ItemInfo which would be a new type

PR #417, opened 2022-04-27
refactor: use iterators in seg hashtable

brayniac:
As noted in #414, we can use iterators to reduce code duplication
within the hashtable module.

This change adds two iterator types to the hashtable, allowing us
to iterate through item info entries for a given hash.

### DIFF
--- a/src/rust/storage/seg/src/hashtable/mod.rs
+++ b/src/rust/storage/seg/src/hashtable/mod.rs
@@ -212,7 +212,7 @@ pub(crate) struct HashTable {
     data: Box<[HashBucket]>,
     started: Instant,
     next_to_chain: u64,
-    _pad: [u8; 8]
+    _pad: [u8; 8],
 }
 
 impl HashTable {
--- a/src/rust/storage/seg/src/hashtable/mod.rs
+++ b/src/rust/storage/seg/src/hashtable/mod.rs
@@ -75,6 +75,7 @@ const MAX_CHAIN_LEN: u64 = 16;
 
 use crate::*;
 use ahash::RandomState;
+use core::marker::PhantomData;
 use core::num::NonZeroU32;
 use metrics::{static_metrics, Counter};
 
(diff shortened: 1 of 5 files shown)


src/rust/storage/seg/src/hashtable/mod.rs
```diff
@@ -109,6 +109,131 @@ pub(crate) struct HashTable {
...
+            return None;
+        }
+
+        let n_item_slot = if self.chain_idx == self.chain_len {
+            N_BUCKET_SLOT
+        } else {
+            N_BUCKET_SLOT - 1
+        };
+
+        let item_info = unsafe { &mut (*self.ptr.add(self.bucket_id)).data[self.item_slot] };
```
hderms, line 217: @brayniac was curious about race conditions w/ HashTable and whether there's some kind of concurrency safe machinery that makes use of unsafe here sensible. Is that accomplished on some higher level via the CAS-related functions I saw elsewhere in this file or through single-threaded execution or something? 

I could probably do better trying to answer this question for myself but I lack familiarity with Segcache

src/rust/storage/seg/src/hashtable/mod.rs
```diff
@@ -109,6 +109,131 @@ pub(crate) struct HashTable {
...
+}
+
+struct IterMut<'a> {
+    ptr: *mut HashBucket,
+    bucket_id: usize,
+    item_slot: usize,
+    chain_len: usize,
+    chain_idx: usize,
+    finished: bool,
+    _marker: std::marker::PhantomData<&'a mut u64>,
```
hderms, line 128: maybe a comment for the intention behind the phantom data would be helpful for future readers

src/rust/storage/seg/src/hashtable/mod.rs
brayniac: The implementation is not designed for concurrent access. We have the storage owned by a single thread. The cas function here allows the memcache cas command to be implemented and doesn't imply concurrency within the server. So it's really just cas for the clients and execution order is serialized in process.

src/rust/storage/seg/src/hashtable/mod.rs
```diff
@@ -109,6 +109,131 @@ pub(crate) struct HashTable {
...
+impl<'a> Iter<'a> {
+    fn new(hashtable: &'a HashTable, hash: u64) -> Self {
+        let bucket_id = (hash & hashtable.mask) as usize;
+        let bucket = hashtable.data[bucket_id];
+        let chain_len = chain_len(bucket.data[0]) as usize;
+
+        Self {
+            hashtable,
+            bucket_id,
+            item_slot: 1,
```
hderms, line 140: yeah I came to a similar conclusion that we should just start on the 1st item as the 0th one is the bucket info, right?

src/rust/storage/seg/src/hashtable/mod.rs
```diff
@@ -162,84 +287,48 @@ impl HashTable {
...
-            bucket = &mut self.data[bucket_id as usize];
         }
 
-        for chain_idx in 0..=chain_len {
-            let n_item_slot = if chain_idx == chain_len {
-                N_BUCKET_SLOT
-            } else {
-                N_BUCKET_SLOT - 1
-            };
+        let rand: u64 = self.rng.gen();
```
hderms, line 302: nice

hderms:
huge improvement to readability, imo

brayniac:
Reminder to myself to do some benchmarking

thinkingfish (approved):

src/rust/storage/seg/src/hashtable/mod.rs
```diff
@@ -541,55 +520,35 @@ impl HashTable {
...
-                            } else {
-                                freq = (freq | 0x80) << FREQ_BIT_SHIFT;
-                            }
-                            if bucket.data[i] == current_info {
-                                bucket.data[i] = (current_info & !FREQ_MASK) | freq;
-                            }
-                        }
-
-                        if cas == get_cas(bucket.data[0]) {
-                            // TODO(bmartin): what is expected on overflow of the cas bits?
```
brayniac, line 579: Good catch @thinkingfish - Per discussion this is fixed as part of #413, comment should be removed

src/rust/storage/seg/src/hashtable/mod.rs
```diff
@@ -541,55 +520,35 @@ impl HashTable {
...
                         }
+                        *item_info = (*item_info & !FREQ_MASK) | freq;
                     }
-                }
-            }
 
-            if chain_idx == chain_len {
-                break;
+                    if cas == get_cas(self.data[bucket_id as usize].data[0]) {
+                        // TODO(bmartin): what is expected on overflow of the cas bits?
```
brayniac, line 544: Same here

PR #418, opened 2022-04-28
feat: #331 Prints the configuration in use for the server and then exits

### LINKED ISSUE #331: segcache-rs: add way to print compile-time constants and configured values
From review of #312: "This reminds me, one of the command line options I had in twemcache/pelikan(C) was to print out all kinds of size constants given the compile-time configuration. This could be a TODO for the future."

See: https://github.com/twitter/pelikan/pull/312#discussion_r647890632

hderms:
debatably fixes #331

I could see this needing to be a bit more self-describing in the future, but this will at least make it easy for the enduser to see the precise configuration in use.

I tried doing this with as minimal investment as possible, hence using Serde/TOML, which seems sensible enough given that that's seemingly the configuration language of choice in the Rust part of Pelikan.

Screenshots:
image
image

### DIFF
--- a/src/rust/config/src/segcache.rs
+++ b/src/rust/config/src/segcache.rs
@@ -93,6 +93,17 @@ impl SegcacheConfig {
     pub fn dlog_interval(&self) -> usize {
         self.dlog_interval
     }
+
+    /// Prints the configuration
+    pub fn print(&self) {
+        let config_toml = self.render_config();
+        println!("Segcache configuration:\n\n{}", config_toml);
+    }
+
+    /// Renders the configuration as a printable string
+    fn render_config(&self) -> String {
+        toml::to_string_pretty(&self).expect("wasn't able to TOML-render config for printing")
+    }
 }
 
 impl AdminConfig for SegcacheConfig {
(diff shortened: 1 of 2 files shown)


brayniac:
Thanks for taking this on. Seems like a good way to handle this. 

Please be sure you accept the CLA / have your account linked. Will take a look at the implementation tomorrow.

hderms:
@brayniac the CLA issue was because I had committed from my work computer which had a different email set up than this user which is the one that created the PR. I rebased/amended the author and now the CLA is passing but it looks like you'll need to re-trigger the workflow

src/rust/config/src/segcache.rs
```diff
@@ -93,6 +93,17 @@ impl SegcacheConfig {
...
         self.dlog_interval
     }
+
+    ///Prints the configuration
+    pub fn print(&self) {
+        let config_toml = self.render_config();
+        println!("Segcache configuration:\n\n{}", config_toml);
+    }
+
+    ///Renders the configuration as a printable string
```
brayniac, line 103: Same here.

src/rust/server/segcache/src/main.rs
```diff
@@ -55,6 +55,11 @@ fn main() {
                 .help("Server configuration file")
                 .index(1),
         )
+        .arg(
+            Arg::with_name("print-config")
+                .help("list & describe all options in config")
```
brayniac, line 60: let's make the help text more consistent with the rest of the help text in the rust implementation. I would prefer "List all options in config", also the option should be `--config` with `-c` as a short form to be consistent with the C version.

brayniac (changes requested):

src/rust/config/src/segcache.rs
```diff
@@ -93,6 +93,17 @@ impl SegcacheConfig {
     pub fn dlog_interval(&self) -> usize {
         self.dlog_interval
     }
+
+    ///Prints the configuration
```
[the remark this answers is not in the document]
hderms: 👍

src/rust/config/src/segcache.rs
hderms: 👍

PR #419, opened 2022-04-28
feat: #368 make admin protocol parsing tolerant of leading/trailing whitespace

### LINKED ISSUE #368: rust: improve admin protocol parsing
Currently the admin protocol parsing is fairly primitive and is less tolerant to extra whitespace than we allow in the memcache protocol implementation.

The admin protocol implementation should be refactored to follow the same patterns used in the memcache protocol implementation. This will allow it to be more robust and extensible.

hderms:
fixes #368 by trimming the command of leading/trailing whitespace before matching

I'm assuming this is what is intended based on reading the C code, but I might be wrong

### DIFF
--- a/src/rust/common/src/bytes.rs
+++ b/src/rust/common/src/bytes.rs
@@ -1,0 +1,37 @@
+// Copyright 2022 Twitter, Inc.
+// Licensed under the Apache License, Version 2.0
+// http://www.apache.org/licenses/LICENSE-2.0
+
+pub trait SliceExtension {
+    ///Trims a type by skipping contiguous whitespace from the beginning and end
+    fn trim(&self) -> &Self;
+}
+
+impl SliceExtension for [u8] {
+    ///Trims an &[u8] by skipping contiguous whitespace from the beginning and end
+    fn trim(&self) -> &[u8] {
+        let first = self.iter().position(|c| !c.is_ascii_whitespace());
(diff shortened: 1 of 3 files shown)


brayniac (changes requested):

brayniac:
I was thinking that we'd do something more like what's in the memcache protocol implementation, eg: https://github.com/twitter/pelikan/blob/master/src/rust/protocol/memcache/src/memcache/wire/request/parse.rs#L73

However, I think the approach here is fine for the simple cases we currently are trying to handle in the admin protocol.

src/rust/common/src/bytes.rs
```diff
@@ -0,0 +1,33 @@
+pub trait SliceExtension {
```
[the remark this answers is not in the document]
hderms: 👍

src/rust/common/src/bytes.rs
```diff
@@ -0,0 +1,37 @@
+// Copyright 2021 Twitter, Inc.
```
brayniac, line 1: 2022 ?

hderms: oh just copied it from another file. I've actually never added a license header and don't really know what the practice is

brayniac: New modules/crates/... should have a license header that reflects the current year.

brayniac (approved):
Looks good to me. Thanks!

PR #425, opened 2022-05-11
Hderms/use new metrics description

hderms:


src/rust/metrics/Cargo.toml
```diff
@@ -15,3 +15,9 @@ common = { path = "../common" }
 macros = { path = "../macros" }
 strum = "0.20.0"
 strum_macros = "0.20.1"
+
+[dependencies.rustcommon-metrics]
```
hderms, line 19: added an import here specifically for the `metrics` crate that now depends on rustcommon-metrics directly instead of through `common`

src/rust/metrics/src/lib.rs
```diff
@@ -5,7 +5,7 @@
 //! A wrapper library for metrics which contains helper functions and macros to
 //! make it easier to use metrics within Pelikan.
 
-pub use common::metrics::{metric, Counter, Gauge, Heatmap, Relaxed};
+pub use rustcommon_metrics::{metric, Counter, Gauge, Heatmap, Relaxed};
```
hderms, line 8: switched this to use the dependency directly and no import error

hderms:
image
Getting this import error when I don't import `rustcommon_metrics` directly i.e. if I leave it to the current behavior of using the re-imported `metrics` from `common` crate

src/rust/metrics/Cargo.toml
brayniac: Can we avoid this? The reasoning for having everything pulled in through common is so that we have just one Cargo.toml to control the versioning of the dependency.

src/rust/core/server/src/threads/admin.rs
```diff
@@ -27,32 +27,53 @@ use std::net::SocketAddr;
...
+#[metric(name="admin_event_loop", crate=common::metrics)]
+static ADMIN_EVENT_LOOP: Counter = Counter::new();
+#[metric(name="admin_event_total", crate=common::metrics)]
+static ADMIN_EVENT_TOTAL: Counter = Counter::new();
+
+#[metric(name="ru_utime", crate=common::metrics)]
+static RU_UTIME: Counter = Counter::new();
+#[metric(name="ru_stime", crate=common::metrics)]
+static RU_STIME: Counter = Counter::new();
+#[metric(name="ru_maxrss", crate=common::metrics)]
```
hderms, line 49: @brayniac not sure if there's a way to avoid this really repetitive `crate = common::metrics` stuff while re-exporting the metrics stuff from `common`

src/rust/metrics/Cargo.toml
hderms: resolved w/ @brayniac offline that I was trying to illustrate an issue between the interplay of re-exporting in the local crate `common` and the `metric` macro from Rustcommon. I didn't want to actually commit this change to master and we're on the same page now. 

By explicitly setting the `crate` attribute, we avoid the previously mentioned issue at the cost of verbosity

hderms:
closing in favor of https://github.com/twitter/pelikan/pull/427

PR #427, opened 2022-05-12
feat: enable metrics descriptions

brayniac:
PR #425 attempted to add metrics description support to Pelikan but
hit some issues due to the way that we have been handling metrics.

This PR adds some new macros to be used in-place of the existing
static_metrics!() macro which allows us to cleanly define metrics
with optional descriptions in the code.

As part of this, metrics are being extensively refactored by moving
the functionality into the 'common' crate.


hderms:
this makes a lot of sense to me. It resolves the crate import/macro issue and the syntax is convenient but ranks low on the magic scale. In the future I wonder if we'd want to introduce these new macros into rustcommon, but we'd have to figure out how to address the crate re-export stuff at that time.

thinkingfish (approved):

PR #440, opened 2022-06-14
Add metric to track epoll_wait returning max events

atn1990:
Add counter for tracking epoll returns max events in the server


src/rust/core/server/src/threads/workers/mod.rs
```diff
@@ -33,6 +33,7 @@ counter!(WORKER_EVENT_TOTAL);
 counter!(WORKER_EVENT_ERROR);
 counter!(WORKER_EVENT_WRITE);
 counter!(WORKER_EVENT_READ);
+counter!(WORKER_EVENT_MAX);
```
brayniac, line 36: Let's add a metric description here. Eg: ```counter!(WORKER_EVENT_MAX, "the number of times the maximum number of events has been returned");```

src/rust/core/server/src/threads/workers/multi.rs
```diff
@@ -118,6 +118,9 @@ where
             }
 
             WORKER_EVENT_TOTAL.add(events.iter().count() as _);
+            if events.iter().count() == self.nevent {
```
brayniac, line 121: instead of iterating twice here, let's have a temporary variable that we can use for both L120 and L121

src/rust/core/server/src/threads/workers/single.rs
```diff
@@ -119,6 +119,9 @@ where
             }
 
             WORKER_EVENT_TOTAL.add(events.iter().count() as _);
+            if events.iter().count() == self.nevent {
```
brayniac, line 122: Same as above

brayniac (changes requested):

src/rust/core/server/src/threads/workers/multi.rs
```diff
@@ -118,6 +118,9 @@ where
...
+                WORKER_EVENT_MAX.increment();
```
thinkingfish, line 122: `WORKER_EVENT_MAX` is ambiguous because it could be interpreted as a gauge value on the max number of events returned in one `poll`. To disambiguate, I suggest renaming this to `WORKER_EVENT_MAX_REACHED` or something to that effect.

thinkingfish:
Once the server change looks good, let's add similar logic to `core/proxy` for parity. Probably both `frontend.rs` and `backend.rs`

src/rust/core/server/src/threads/workers/mod.rs
atn1990: Done

src/rust/core/server/src/threads/workers/single.rs
atn1990: done

src/rust/core/server/src/threads/workers/multi.rs
atn1990: done

src/rust/core/server/src/threads/workers/multi.rs
atn1990: done

PR #442, opened 2022-06-14
reorganize project

brayniac:
Reorganizes the Pelikan repo to promote the Rust codebase to be top-level for this project. The C code remains in the `legacy` folder as a reference. Top-level readme now focuses entirely on the Rust code, and the legacy readme focuses on the legacy code.


thinkingfish:
The only thing I'm concerned about is if someone has a relatively big change against the original layout, and is now in need of a massive merge conflict exercise. I wonder if there's any good way of getting that info out of examine all the public forks.

thinkingfish (approved):

PR #452, opened 2022-07-28
Some update for JSegcache

JunchengYTwitter:
This PR contains multiple things (let me know if you want to split them)
1. Add `get_age` in segments, then add a new field `age` in `Item` so that we can return age to caller. 
2. Add `RichItem` to return `item_info` and `item_info_ptr`, these two are used to support opportunistic concurrency control. Not sure adding a new struct is better than adding the fields to `Item`, but given there might be fields (like these) that we often do not need, creating a new struct might be better. 
3. add a call to expire before we evict, this avoids the problem that users may forget to call `expire`.

### DIFF
--- a/src/storage/seg/src/hashtable/mod.rs
+++ b/src/storage/seg/src/hashtable/mod.rs
@@ -405,7 +405,6 @@ impl HashTable {
                         item_info,
                         get_cas(self.data[(hash & self.mask) as usize].data[0]),
                     );
-                    assert!(item.verify_hashtable_entry()); 
                     item.check_magic();
 
                     return Some(item);
--- a/src/storage/seg/src/hashtable/mod.rs
+++ b/src/storage/seg/src/hashtable/mod.rs
@@ -393,9 +393,11 @@ impl HashTable {
                             freq = (freq | 0x80) << FREQ_BIT_SHIFT;
                         }
                         // TODO: this needs to be atomic
+                        // worse case new item insert fails
                         *item_info = (*item_info & !FREQ_MASK) | freq;
                     }
 
+                    println!("{} == {}", item_info_val, *item_info);
                     let age = segments.get_age(item_info_val).unwrap();
                     let item = RichItem::new(
                         current_item,
@@ -404,6 +406,7 @@ impl HashTable {
                         item_info,
                         get_cas(self.data[(hash & self.mask) as usize].data[0]),
                     );
+                    assert!(item.verify_hashtable_entry()); 
                     item.check_magic();
 
                     return Some(item);
--- a/src/storage/seg/src/hashtable/mod.rs
+++ b/src/storage/seg/src/hashtable/mod.rs
@@ -397,12 +397,11 @@ impl HashTable {
                         *item_info = (*item_info & !FREQ_MASK) | freq;
                     }
 
-                    println!("{} == {}", item_info_val, *item_info);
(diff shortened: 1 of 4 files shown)


brayniac:
@JunchengYTwitter - this looks good overall. Just needs rustfmt. I'm okay with this remaining a single PR.

JunchengYTwitter:
updated

src/storage/seg/src/item/mod.rs
```diff
@@ -83,6 +89,99 @@ impl std::fmt::Debug for Item {
     }
 }
 
+/// Items are the base unit of data stored within the cache.
+pub struct RichItem {
```
brayniac, line 93: Is there a reason this can't be:
```pub struct RichItem {
  item: Item,
  item_info: u64,
  item_info_ptr: *const u64,
}```

Comment (JunchengYTwitter) [reply src/storage/seg/src/item/mod.rs:93]: nope, your proposal is better, I can change it :)

File: src/storage/seg/src/item/mod.rs [hunk 2]
```diff
@@ -83,6 +89,99 @@ impl std::fmt::Debug for Item {
...
+    }
+}
+
+impl std::fmt::Debug for RichItem {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::result::Result<(), std::fmt::Error> {
+        f.debug_struct("Item")
+            .field("cas", &self.cas())
+            .field("raw", &self.raw)
+            .field("item_info", &self.item_info)
+            .field("item_info_ptr", &self.item_info_ptr)
```
brayniac, line 180: Doesn't this print the ptr address? Is that really useful?

JunchengYTwitter: yes, it is not useful, printing the value itself is not very useful as well. I can print all three (two value, one pointer) or we can remove the print. What do you think?

src/storage/seg/src/hashtable/mod.rs
```diff
@@ -308,8 +308,106 @@ impl HashTable {
...
+                    // update item frequency
+                    let mut freq = get_freq(*item_info);
+                    if freq < 127 {
+                        let rand = thread_rng().gen::<u64>();
+                        if freq <= 16 || rand % freq == 0 {
+                            freq = ((freq + 1) | 0x80) << FREQ_BIT_SHIFT;
+                        } else {
+                            freq = (freq | 0x80) << FREQ_BIT_SHIFT;
+                        }
+                        // TODO: this needs to be atomic
```
brayniac, line 400: I'm having trouble understanding why - we hold a mutable reference to the hashtable here - so there can be no other read or write references.

src/storage/seg/src/item/mod.rs
brayniac: It seems like printing if it's stale or not would maybe have value here?

src/storage/seg/src/item/mod.rs
```diff
@@ -83,6 +89,99 @@ impl std::fmt::Debug for Item {
...
+    }
+
+    pub fn age(&self) -> u32 {
+        self.age
+    }
+
+    // used to support multi readers and single writer
+    // return true, if the item is evicted/updated since being
+    // read from the hash table
+    pub fn is_not_changed(&self) -> bool {
```
brayniac, line 152: i don't love the `is_not_` here - can this just be `is_current()`?

PR #454, opened 2022-08-17
momento-proxy: add RESP support

brayniac:
Adds basic RESP (Redis Serialization Protocol) support to the
Momento proxy. This allows standard redis clients to communicate
with the Momento caching service for basic get/set requests.


src/protocol/resp/src/lib.rs
```diff
@@ -2,8 +2,11 @@
 // Licensed under the Apache License, Version 2.0
 // http://www.apache.org/licenses/LICENSE-2.0
 
-mod get;
-mod set;
+mod request;
```
thinkingfish, line 5: The basic types of RESP are used to represent both request and response.

I think the code structure should reflect this relationship, where `protocol/resp` should use one (sub-)module to represent primitive types and their serde operations; and the other module to represent commands and their serde, implemented on top of the primitive types / operations.

src/proxy/momento/src/protocol/redis/mod.rs
```diff
@@ -0,0 +1,5 @@
+mod get;
```
thinkingfish, line 1: Is it intentional to name this module `redis` instead of `resp`?

src/proxy/momento/src/protocol/redis/get.rs
```diff
@@ -0,0 +1,105 @@
...
+pub async fn get(
+    client: &mut SimpleCacheClient,
+    cache_name: &str,
+    socket: &mut tokio::net::TcpStream,
+    key: &[u8],
+) -> Result<(), Error> {
+    GET.increment();
+
+    // check if any of the key is invalid
+    if std::str::from_utf8(key).is_err() {
```
thinkingfish, line 15: The response construction seems a bit too unstructured, e.g. there isn't a simple way to enumerate all the possible error messages for `get` handling. 

I think with some refactoring of the basic protocol parser, we can add some helper traits to give request-response processing more structure. E.g. we can then declare an `enum` type that lists all the possible exceptions of handling `get` request, and provide a trait in protocol mod to write an error message with customizable details, etc.

src/protocol/resp/src/lib.rs
thinkingfish: It's true current redis commands are all multistrings, but I think if we want to expand it (e.g. allowing options/maps), we will make full use of all the types in both requests and responses.

brayniac: Makes sense. This sounds good.

src/proxy/momento/src/protocol/redis/get.rs
brayniac: What would you feel about tackling this after the initial merge? I agree that a more detailed error type within the protocol layer makes sense. `thiserror` would be a good choice to get some nice errors w/ pretty conversion to strings for general case and still would give us enum errors for use-cases where we need to do some more custom handling.

At this particular spot, we need to make sure the key is a utf8 string, which is not strictly a requirement for RESP or Memcache protocol (which allow byte strings for keys) but is a requirement when using Momento. I don't think this particular case would be handled by the error type within the protocol crates, as it's specific to the transformation from a RESP request and the Momento GRPC request. These types of transforms should eventually be handled more generally so we can produce proxies which speak different protocols on the frontend and backend.

src/proxy/momento/src/protocol/redis/mod.rs
brayniac: Fixed

src/protocol/resp/src/lib.rs
brayniac: Take a look when you get a chance.

src/proxy/momento/src/protocol/redis/get.rs
thinkingfish: makes sense

src/protocol/resp/src/message/integer.rs
```diff
@@ -0,0 +1,42 @@
+// Copyright 2022 Twitter, Inc.
+// Licensed under the Apache License, Version 2.0
+// http://www.apache.org/licenses/LICENSE-2.0
+
+use super::*;
+
+#[derive(Debug, PartialEq, Eq)]
+pub struct Integer {
+    pub(crate) inner: u64,
```
kevyang, line 9: i think RESP integers are signed 64 bit, not unsigned, according to the protcol spec https://redis.io/docs/reference/protocol-spec/

brayniac: You're right. Fixed

thinkingfish (approved):

kevyang (approved):
lgtm. i think some of the unit test coverage could be more complete but we can always add that in at a later time.

src/protocol/resp/src/message/simple_string.rs
```diff
@@ -0,0 +1,47 @@
...
+        },
+    ))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn parse() {
```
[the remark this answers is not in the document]
brayniac: Strings in the RESP format are binary safe and should be able to include a CRLF. The length of the string is provided in the serialized form. For inline request, if wrapped in double quotes, the string is also binary safe and may include CRLF.

I think we should get this merged and create an issue to add more unit tests and improve some of the error handling (per Yao's feedback).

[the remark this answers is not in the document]
kevyang: agreed! I will merge it then. thanks!

PR #459, opened 2022-09-16
WIP: segcache: use mutex and jemalloc

brayniac:
Draft PR, needs performance evaluation

Changes threading model to use shared storage, protected
by a mutex. This removes the overhead of cross-thread
signaling and communication on the request path in favor of
using coarse-locking.

Sets the global allocator to jemalloc for pelikan_segcache_rs


brayniac:
For a 9 core configuration on a Intel(R) Xeon(R) Platinum 8352V CPU @ 2.10GHz w/ 100GbE NIC (E810-C) (dual-socket machine, NIC queues reduced to one per physical core, SMT enabled, threads pinned to socket 0 (with NIC) but avoid the HW threads that handle NIC IRQs) this change brings the redline from 728kqps -> 830kqps for a 80:20 Read:Write workload

PR #463, opened 2022-09-28
ci: set opt-level for debug builds

brayniac:
Sets the opt-level to 3 for debug builds in an attempt to reduce flakeyness of integration tests (which cannot be reproduced outside of the ci environment).


thinkingfish:
I've opened #464 so we don't forget about this bug. With that, we can merge this short-term fix.

brayniac:
closing in favor of #465 which should make the tests more reliable.

PR #465, opened 2022-09-29
ci: rework integration tests for segcache

brayniac:
Reworks the integration test for segcache so that multi-step test cases run over the same connection to ensure proper sequencing.


thinkingfish (approved):
the cas related tests are not super robust, but I think it's OK for now. To actually get the cas value back to be used by cas would require more change to how the test function is implemented.

PR #468, opened 2022-09-30
Introduce pelikan_bloomcache

swlynch99:
This is code-complete and lightly tested. I haven't tested this beyond some basic smoke tests but ultimately the code is pretty simple.

This is a draft PR so that people can look at it while I split things up and do some performance testing.

PR #469, opened 2022-10-03
Implement HTTP protocol

swlynch99:
This PR introduces a rather simple HTTP protocol. It supports 3 commands - `GET`, `PUT` (set), and `DELETE` which is enough for simple cache use-cases. I have included tests for both request parsing and response emission.

In detail, this PR includes:
- A HTTP parser that wraps [`httparse`](https://docs.rs/httparse)
- A very simple HTTP response builder that does no validation
- Tests for the above

Due to limitations around what pelikan allows parsers to do (they can't emit responses) the response parser doesn't directly return a request object but instead returns a wrapper around a `Result<Request, Response>`. The error response should be sent by the client in the place of creating its own response.

This PR is the first part of breaking apart #468 into mergeable chunks.


src/protocol/http/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "protocol-http"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+arrayvec = "0.7.2"
```
brayniac, line 7: I think we should move the rest of these versions up to the workspace.

src/protocol/http/src/lib.rs
```diff
@@ -0,0 +1,27 @@
+// Copyright 2022 Twitter, Inc.
+// Licensed under the Apache License, Version 2.0
+// http://www.apache.org/licenses/LICENSE-2.0
+
+//!
```
brayniac, line 5: Are you planning to add some docs here before mereg?

src/protocol/http/tests/request.rs
```diff
@@ -0,0 +1,104 @@
+use assert_matches::assert_matches;
```
brayniac, line 1: license header.

src/protocol/http/src/lib.rs
swlynch99: Yeah - forgot about that. Will add some.

brayniac (approved):
LGTM

PR #478, opened 2022-10-19
fix clippy warnings for github/pelikan/src/protocol/resp lib

0xcdc:
removed some assignments to _
switchted from loop with an if break to a while 
added Eq to TimeType in the expiry.rs common code section

all tests from the root of pelikan pass


brayniac (approved):
Looks good. Thanks for addressing these!

PR #480, opened 2022-10-19
ci: split macos from linux builds

brayniac:
macos builds continue to be flakey in ci for debug builds. In order to make ci less flakey, this change splits linux and macos builds and drops debug mode for macos builds.

### DIFF
--- a/.github/workflows/cargo.yml
+++ b/.github/workflows/cargo.yml
@@ -9,11 +9,53 @@ env:
   RUST_BACKTRACE: full
 
 jobs:
-  build:
+  build-linux:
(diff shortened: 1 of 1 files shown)


swlynch99 (approved):
The change itself LGTM but the macos-release job is failing for what looks like a similar reason as the debug one.

brayniac:
> The change itself LGTM but the macos-release job is failing for what looks like a similar reason as the debug one.

Yeah - well, at least we can rule out debug mode as part of it. I'll open a separate PR to start troubleshooting this

PR #486, opened 2022-10-26
badd command second attempt

0xcdc:


brayniac:
Code all looks fine to me now.

Tagging @thinkingfish to review and weigh-in on how we want to handle the btree family of commands in the resp protocol crate.

src/proxy/momento/src/frontend.rs
```diff
@@ -86,6 +86,7 @@ pub(crate) async fn handle_resp_client(
                 let request = request.into_inner();
 
                 match request {
+                    resp::Request::BAdd(_) => todo!(),
```
brayniac, line 89: Actually, just noticed this - this introduces a panic path in a service. We should have a more defensive approach and treat it as an invalid command. I think it'd be acceptable to add a 
```_ => {
                  println!("bad request");
                  let _ = socket.write_all(b"CLIENT_ERROR\r\n").await;
                  break;
              }```

To the end of this match to catch any commands that aren't implemented.

**Review by brayniac — CHANGES_REQUESTED** [review]

File: src/protocol/resp/src/request/badd.rs [hunk 2]
```diff
@@ -0,0 +1,142 @@
...
+
+use super::*;
+use std::io::{Error, ErrorKind};
+use std::sync::Arc;
+
+type ArcByteSlice = Arc<Box<[u8]>>;
+type ArcKeyValuePair = (ArcByteSlice, ArcByteSlice);
+
+#[derive(Debug, PartialEq, Eq)]
+pub struct BAddRequest {
```
brayniac, line 13: Since `badd` command is not a upstream redis command, it would be good to document what this is. Adding some rustdoc for the struct would be helpful

src/protocol/resp/src/request/mod.rs
```diff
@@ -87,6 +89,9 @@ impl Parse<Request> for RequestParser {
 
                 match &array[0] {
                     Message::BulkString(c) => match c.inner.as_ref().map(|v| v.as_ref().as_ref()) {
+                        Some(b"badd") | Some(b"BADD") => {
```
brayniac, line 92: In the future we may consider using conditional compilation to enable/disable groups of commands. For now this is probably okay as-is.

src/protocol/resp/src/request/badd.rs
```diff
@@ -0,0 +1,144 @@
...
+// http://www.apache.org/licenses/LICENSE-2.0
+
+use super::*;
+use std::io::{Error, ErrorKind};
+use std::sync::Arc;
+
+type ArcByteSlice = Arc<Box<[u8]>>;
+type ArcKeyValuePair = (ArcByteSlice, ArcByteSlice);
+
+/// Represents the btree add command which was added to Twitter's internal version of redis32.
```
brayniac, line 12: Keep doc lines to 80 characters width maximum to match the rest of the codebase. Split this across multiple-lines and it'll be fine.

brayniac (approved):

PR #494, opened 2023-10-11
feat: add nvme ssd cache layer

linzs148:


1a1a11a:
Hi @linzs148, interesting work! However, the development of pelikan and Segcache are no longer at Twitter, are you interested in some open-source work to support flash in Segcache?

